Issue #5, opened 2022-06-10
Setup Initial CI/CD


RNKuhns:
Setup the package's initial CI/CD routine, to include:

* Automated code style and quality linting
* Testing (and test coverage reporting)
* Documentation building
* Releasing the package

PR #8, opened 2022-06-16
Add CI Routines

RNKuhns:
#### Reference Issues/PRs

Fixes the setup of an initial CI/CD routine as outlined in issue #5.

Adds configuration files and workflows necessary for continuous integration and continuous deployment, including:

- Continuous integration
    -  Add pre-commit config file so that linting can be performed automatically
    -  Add config files for linters
    -  Add Github Actions to automatically run all of the package's unit tests

No new dependencies are added.

In several cases, I've made alterations to the CI routine used by sktime. Differences include:

- Including Python 3.10 in the test workflow
- Standardizing on using the conda-incubator/setup-miniconda@v2 to setup Python on all operating systems
- using pre-commit.ci in addition to code quality workflow based on pre-commit GHA (which won't be updated any longer)
   - pre-commit.ci is faster than the action, and has additional features (like auto-update your PRs based on code formatting fixes!)


RNKuhns:
@fkiraly I think I've got the CI portion underway. I'm tryign to remember from when I set this up before on my own repos, is there a way to turn the CI on so that it runs on the pull request that commits it? (The things I read online suggest that it runs after commit)

RNKuhns:
I'd also like to move from using the pre-commit action (pre-commit/action@v2.0.0) to pre-commit.ci. The [pre-commit action](https://github.com/pre-commit/action) is in maintenance only mode. I've used pre-commit.ci and can attest to it being faster. It also automatically commits the linting errors that it can fix for you!

The downside is that I haven't yet figured out how to have a failur of the pre-commit.ci run of the pre-commit hooks cancel the test workflows. 

Even if we don't get that figured out in the short-run, my thoughts on it are that this might matter more on a package like sktime where the tests can run for a while. With BaseObject, we'd expect our tests to run very quickly, so the amount of github action time we waste is probably pretty small. Also, outside of our initial refactoring, I'm guessing that we won't have heavy development on this (it being general base class functionality, you'd want it to change slowly).

RNKuhns:
@fkiraly the unit tests are failing because the actual package (baseobject._base.py) is importing functionality from sktime locations. That means `from baseobject import BaseObject` generates and error.

We need to figure out a few of those things so we can get the unit tests up and running.

fkiraly (approved):
Looks good!

Question: can we do precommit not via GHA? That would allow us to not depend on travis.

PR #9, opened 2022-06-17
Create dependency-review.yml

RNKuhns:
I'm adding an intial github action (GHA), since the documentation for GHA suggests that you have to add an action to a repository for them to start running. 

This is also useful so that we can easily make sure that we don't introduce any dependencies with known high or critical security vulnerabilities.


RNKuhns:
I've combined this pull request into the main CI pull request.

PR #11, opened 2022-06-24
[ENH] lookup utility `all_estimators`

fkiraly:
This PR moves a slightly refactored lookup utility `all_estimators` over from `sktime`.

Changes are made to decouple the utility from:
* the base class registry. Instead, this is passed as an alias dictionary `class_dict`. Estimator types are now passed as classes directly, and can be passed as strings (like in `sktime`) if the alias dictionary is provided.
* `pandas`. This is now a soft dependency that is needed only when `as_dataframe` is set to `True`, allowing the `baseobject` package to not depend on `pandas`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,3 +2,9 @@
 # Welcome to BaseObject
 
 > A base class for scikit-learn-like parametric objects
+
+Temporary location of refactoring the `BaseObject` class and related functionality from `sktime` into a general package for unified parametric object interfaces.
+
+Maintainers:
+@rnkuhns (project lead)
+@fkiraly
--- a/baseobject/_lookup.py
+++ b/baseobject/_lookup.py
@@ -1,17 +1,19 @@
 # -*- coding: utf-8 -*-
 # copyright: BaseObject developers, BSD-3-Clause License (see LICENSE file)
-"""
-Registry lookup methods.
+"""Tools to lookup information on code artifacts in a Python package or module.
 
 This module exports the following methods for registry lookup:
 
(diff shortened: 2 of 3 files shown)


fkiraly:
why is this failing, not sure I understand what is going on

RNKuhns:
@fkiraly I've got the `package_metadata` function added. It walks a package to get high-level info on included functions and classes.

Right now this function is requiring users to supply a package/module name and a path. However, we can get the path from the package/module (import it in the function to retrieve the path, then walk the path). I think that will cover our primary use-case of being able to return metadata on an installed package. For someone contributing to the package, they will have a development (editable) install and this will work. During the CI runs, the package is checked out and setup at the start so this should also work (including tests will verify). 

If that sounds reasonable, I'll push an update that removes the need for users to specify a `path` parameter in `package_metadata`.

RNKuhns:
CI errors are related to a type issue (using Iterable from collections.abc versus typing depending) and Python versions.

fkiraly:
note to myself: review @RNKuhns´s diff on `all_estimators`

fkiraly (commented):
I cannot formally approve since you added diff to my PR, but I approve the diff (which is massive).

Primarily, my concern is that the signature of `all_objects` is a superset of functionality in `sktime` `all_estimators`, which is satisfied.

We will also need tests for this, but at the moment not sure what the best setup for that is. We would need to set up an entire test module which has a few classes, right? I would defer that to separate PR.

PR #14, opened 2022-07-01
[ENH] up `flake8` max line length to 88

fkiraly:
This increases max line length for `flake8` to 88, in line with `sktime`.


RNKuhns:
@fkiraly the linter configs are included in #8. It was a pain without them in the repository (lots of false positives generated locally without the linter configs), but they should be included now.

PR #16, opened 2022-07-01
[ENH] test framework

fkiraly:
This moves the test framework from `sktime`, currently only for `BaseObject` features.

This means:
* no scenarios
* no "state changing methods"

Major changes:
* added `TestAllObjects` class for testing. In the future, we will probably want to separate the "original" class and showcase how users can inherit. For now, this is running from its original directory
* moved some utilities from `sktime` into `testing.utils`
* for code from `sktime` removed dependencies on inner `sktime` code and hard dependencies on `pandas`
* `mock_package` that we can later extend, used to showcase and test `TestAllObjects`

Minor changes:
* in `_lookup`, changed "estimator" terminology to "object" terminology (docstrings, variable names)


fkiraly:
I'm not quite sure what's happening with the linter here, can you help me, @RNKuhns?

It does not like the `assert` statements in the tests, but they should be able to live there?
Also, the `X` etc which are common usage in the sklearn compatible world

fkiraly:
& you changed the GHA yml files, they seem to be broken?
Let's talk today about how to fix this, I need to be able to work on the tests.

fkiraly:
@RNKuhns, I think this is ready, at least in first (very experimental) instance

baseobject/_lookup.py
```diff
@@ -584,17 +584,14 @@ def all_objects(
...
-    estimator_types: class or list of classes, default=None
-
+    object_types: class or list of classes, default=None
         - If class_lookup is provided, can also be str or list of str
           which kind of objects should be returned.
         - If None, no filter is applied and all estimators are returned.
         - If class or list of class, estimators are filtered to inherit from
           one of these.
         - If str or list of str, classes ca be aliased by strings, via class_lookup.
-
```
RNKuhns, line 595: I think we might need the spaces added back here to render the bulleted lists correctly. But I'll iterate with that in the doc PR.

baseobject/mock_package/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
```
RNKuhns, line 1: Eventually I'd like for us to come up with somewhere else for this. But as we've talked about, the best location is not clear yet. So let's keep it here in V0.1.0 and revisit in V0.2.0.

One option is to move it as a submodule under "tests", since it isn't really user functionality. But as you've pointed out there may be some documentation/example benefit of it. So I think maybe a better option is to have a "template" mock package in its own repository that we can use in the tests and point to as an example of how you can use BaseObject.

baseobject/testing/test_all_objects.py
```diff
@@ -0,0 +1,789 @@
...
+    #   if this were not executed, object_instance would keep state changes
+    #   within executions of the same test with different parameters
+    @pytest.fixture(scope="function")
+    def object_instance(self, request):
+        """object_instance fixture definition for indirect use."""
+        # esetimator_instance is cloned at the start of every test
+        return request.param.clone()
+
+
+class QuickTester:
```
RNKuhns, line 230: Are you flexible on the name? Maybe TesterMixin? I'm going to approve this, but maybe we include a quick PR that updates the name before V0.1.0 release?

RNKuhns: Mainly I'm thinking that `QuickTester` implies that the primary benefit is speed or that this was done as a quick/intermediate output (but goal of package is well thought out testing interface, so we know that is not the case)

RNKuhns (approved):
@fkiraly I agree with everything with only two minor things I'd like to see change, but I'm defering on both for now.

1. Is that I think we should eventually move the mock_package somewhere else.
2. I think we should rename QuickTester. But that can be a quick follow-on PR if we have time before V0.1.0

baseobject/testing/utils/deep_equals.py
```diff
@@ -0,0 +1,308 @@
...
+
+__all__ = ["deep_equals"]
+
+
+import numpy as np
+
+from baseobject.testing.utils._dependencies import _check_soft_dependencies
+
+
+def deep_equals(x, y, return_msg=False):
```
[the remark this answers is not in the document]
RNKuhns: As we've talked about, that would be a nice outcome, but is a different PR later on.

Issue #24, opened 2022-07-09
Investigate speed up of CI Test Routine


RNKuhns:
**Is your feature request related to a problem? Please describe.**
It may be possible to speed up the Ci test routine by speeding up the use of conda. Potential options to investigate include:

1. Use cached Conda environments. See section on [caching](https://github.com/marketplace/actions/setup-miniconda) for setup-miniconda action
2. Another option is to investigate the use of `mamba` from the [setup-miniconda action](https://github.com/marketplace/actions/setup-miniconda)

PR #33, opened 2022-07-29
caching environment for test routine

KatieBuc:
#### Reference Issues/PRs
Addressing issue #24, uses a cashed conda environment

Following the example documentation given here: https://github.com/marketplace/actions/setup-miniconda

The steps are:
- setup a Miniconda variant
- adds current date to the cache key (refreshes cache every 24 hours)
- if the cache does not exist, update the environment based on the environment file 

"conda (and mamba) resolves the dependencies declared in the YAML file according to the packages available on the channels at installation time. Since packages are updated all the time, you will not see these changes reflected in the cache until the key gets updated by date.

This means that the same environment file can make your tests pass locally but fail on CI, or the other way around. In that case, reset the cache manually to see if that leads to consistent results, or use a resolved environment file."


fkiraly:
superseded by new GHA setup

PR #36, opened 2022-08-07
[ENH] robustify `all_objects` exclusion behaviour

fkiraly:
This robustifies the `all_objects` exclusion behaviour, by avoiding to import excluded modules or their sub-modules on the `ignore_modules` list.

This is in particular relevant if such an import would lead to circular imports.

The change in logic is to avoid importing the ignored modules altogether, instead of importing and then discarding - the former avoids circular imports, the latter runs into the problem of circular imports, potentially.

Also see https://github.com/alan-turing-institute/sktime/pull/3198 in `sktime`.

### DIFF
--- a/baseobject/_lookup.py
+++ b/baseobject/_lookup.py
@@ -690,10 +690,6 @@ def all_objects(
     def _is_private_module(module):
         return "._" in module
 
-    def _is_ignored_module(module):
-        module_parts = module.split(".")
-        return any(part in modules_to_ignore for part in module_parts)
-
     def _is_base_class(name):
         return name.startswith("_") or name.startswith("Base")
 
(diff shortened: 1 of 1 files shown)


RNKuhns:
@fkiraly I made some minor changes to your updates. I updated the `package_metadata` function to use the new `_walk` functionality. This meant factoring it outside of `all_objects`. Also finished adding params to the `package_metadata` function that I had left out before.

fkiraly:
in-principle happy with the changes.

Side observation, I think the docstring for `package_metadata` should explain the structure of the return dict properly, and maybe also give the function a better name? So it's clear that it retrieves metadata, rather than *being* metadata. But that observation is not specific to this PR, since it is not related to a change.

RNKuhns:
> in-principle happy with the changes.
> 
> Side observation, I think the docstring for `package_metadata` should explain the structure of the return dict properly, and maybe also give the function a better name? So it's clear that it retrieves metadata, rather than _being_ metadata. But that observation is not specific to this PR, since it is not related to a change.

@fkiraly your suggestions make sense to me. I've updated the docstring and renamed `package_metadata` as `gather_package_metdata`.

PR #38, opened 2022-08-26
Update readme to include CI status and basic content

RNKuhns:
This PR provides a quick blurb about the package, a table of badges showing the status of our CI (among other things), and how to install the package from PyPi via pip.

We can always add more later, but this gives us some basic info on the README ahead of v0.1.0 release.


RNKuhns:
@fkiraly I believe this should be ready for your review.

Note that in the installation section, I mention Python versions through 3.10, but the bade doesn't say 3.10 yet. Should be cleared up when we release v0.1.0.

fkiraly (approved):
good

PR #53, opened 2022-10-07
Fix CI error from test coverage reference to old package name

RNKuhns:
This fixes an error that is causing test failures in other PRs due to a reference to the baseobject package in the configuration for test coverage in the package's `pyproject.toml`. The reference to `baseobject` was updated to `skbase`. 

Also updated the excluded directories to exclude the docs and tests from coverage.

None.

Tests run successfully and decision on excluding tests and docs from coverage.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,7 @@ convention = "numpy"
 show_missing = true
 
 [tool.coverage.run]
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Thanks!

PR #58, opened 2022-10-12
Post Release Documentation Fixes

### LINKED ISSUE #54: [DOC] Add sphinx-copybutton extension
#### Describe the issue linked to the documentation

The [sphinx-copybutton](https://sphinx-copybutton.readthedocs.io/en/latest/index.html) package provides a nice sphinx extension to make it easy to copy code blocks.

#### Suggest a potential alternative/fix

We should add this. It will make it easier to copy the code blocks in the install instructions now, and all the code examples later.

RNKuhns:
This PR includes some documentation improvements that became apparent after the initial package release. This includes fixes to update the project's sphinx config file, so it correctly lists the project as `skbase`.

Other minor changes include:

- Minor tweaks to the `pydata-sphinx-theme` configuration including, adding a light/dark theme toggle and icon link (at end of top navbar) to pypi.
- Dynamically setting the copyright year
- Getting version by importing `skbase` instead of hardcoding it in `conf.py`
- Adding sphinx-copybutton extension that fixes #54 
- Fixes 3 typos (user guide heading and 2 references on get started page) that had dash instead of apostrophe when trying to reference multiple `BaseObjects` but typed `BaseObject-s`

No.

Does the documentation configuration look correct? Other than the 3 typos, no changes to documentation content are included. Any doc content changes will be handled in separate PRs.

#### Any other comments?
None.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
       name: isort
 
 - repo: https://github.com/psf/black
-  rev: 22.8.0
+  rev: 22.10.0
   hooks:
   - id: black
     language_version: python3
--- a/docs/source/_static/switcher.json
+++ b/docs/source/_static/switcher.json
@@ -1,12 +1,17 @@
 [
-    {
+  {
       "name": "dev",
       "version": "latest",
       "url": "https://baseobject.readthedocs.io/en/latest/"
     },
     {
-      "name": "0.1.0",
+      "name": "0.2.0 (stable)",
       "version": "stable",
       "url": "https://baseobject.readthedocs.io/en/stable/"
+    },
+    {
+      "name": "0.1.0",
+      "version": "v0.1.0",
+      "url": "https://baseobject.readthedocs.io/en/v0.1.0/"
     }
 ]
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -1,8 +1,8 @@
 .. _api_ref:
 
-===
-API
-===
+=============
+API Reference
+=============
 
 Welcome to the API reference for ``skbase``.
 
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -1,8 +1,8 @@
 .. _api_ref:
 
-=============
-API Reference
-=============
+===
+API
+===
 
 Welcome to the API reference for ``skbase``.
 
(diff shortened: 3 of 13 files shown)


RNKuhns:
@fkiraly can you take a quick look at this PR. Only tricky part was figuring out how to let us import skbase in the conf.py without having a CI error.

Just importing skbase locally (assuming editable install) would let you run pytest without errors. But in CI it would lead to collection errors (because it was finding the tests but not from expected "locations"). Excluding the docs folder from pytest solves the problem.

fkiraly:
> Just importing skbase locally (assuming editable install) would let you run pytest without errors. But in CI it would lead to collection errors (because it was finding the tests but not from expected "locations"). Excluding the docs folder from pytest solves the problem.

Do you have an explanation for that?

RNKuhns:
> Looks simple enough, great!
> 
> * I would leave "API reference" as such, since API feels more vague/unspeific.
> * is there not a way to automatically update the switcher?
> * regarding `BaseObjects`, I think the in-line code formatting should apply only to object names. I.e., it would be `BaseObject`-s and not `BaseObjects`, because the object is called `BaseObject` and there isn't a class called `BaseObjects`

- I'm fine reverting to "API Reference". 
- Unfortunately, there isn't a built-in way to automatically update the switcher. Though the update is formulaic (increment the name of the switcher item pointing to the stable release to the newest release version, and add an item that points to the prior version's docs (which was previously accessible through stable). We might be able to automate that later on, but will need to remember to increment things as part of our release process in the short-run.
- I'm fine using ``BaseObject``-s (even if I don't like it visually).

RNKuhns:
> > Just importing skbase locally (assuming editable install) would let you run pytest without errors. But in CI it would lead to collection errors (because it was finding the tests but not from expected "locations"). Excluding the docs folder from pytest solves the problem.
> 
> Do you have an explanation for that?

I wish I had a good explanation for why this happens. But I tested and confirmed it is the import of skbase to dynamically determine the version that caused the issue. After going around in a circle for a while, I compared the config with sktime and noticed the difference I've changed here, and it fixed it.

RNKuhns:
@fkiraly I think I've addressed your comments, let me know if this is okay to merge.

fkiraly (approved):
Happy with this in-principle!

As said, the switcher update makes the process more brittle (easy to forget that step), but I do not have a strong opinion.

RNKuhns:
@fkiraly I like the idea of the switcher (being able to see docs for different versions). I'm willing to make step-by-step instructions for the release procedures so we don't forget to make that update.

PR #62, opened 2022-10-23
Add tests for BaseObject and BaseEstimator

### LINKED ISSUE #61: Add Tests for BaseEstimator
Add tests of BaseEstimator.

Should test that `BaseEstimator` inherits from `BaseObject`. Should also test the presence of the interface (`is_fitted`, `check_is_fitted`) and that it works correctly.

### LINKED ISSUE #63: Test additional BaseObject functionality
We test the primary interfaces provided by `BaseObject` but we still need to test some additional functionality. Running list is provided below.

We need to add tests of the following:
- `BaseObject._get_init_signature()` is `object.__init__`
- `BaseObject_get_init_signature()` raises error for `VAR_POSITIONAL`
- `BaseObject.get_param_names()`
- `BaseObject.set_params()` with no params returns original object.
- `BaseObject.set_params()` raises error when param is not a valid params.
- `BaseObject.set_tags()` tag setting when underlying object doesn't have _tags_dynamic attribute (do we want to add it like we currently do (then we add test class) or do we want this to raise error (since this would imply the subclass doesn't correctly inherit the entire`BaseObejct` interface)?
- `BaseObject.clone_tags()`
- `BaseObject.get_test_params()`
- `BaseObejct.create_test_instance()`:
- `BaseObject.create_test_instances_and_names()`
- `BaseObject._has_implementation_of()`
-  `BaseObject._components()` needs to test correct error raised if base_class is not a class or it is a class that isn't a subclass of `BaseObject`.s returns original object.

**Describe alternatives you've considered**
Add tests to covering these areas to `test_base.py`.

RNKuhns:
#### Reference Issues/PRs
Fixes #61 and fixes #63.

This adds additional tests of the `BaseEstimator` interface as described in #61 and additional tests of the `BaseObject` interface as described in #63.

No new dependencies.

Does this provide desired testing of interface.


RNKuhns:
@fkiraly I'm making good progress towards adding more tests of our base functionality. 

Just plan to add tests for the following yet as part of this PR:
- BaseObject.get_test_params()
- BaseObejct.create_test_instance():
- BaseObject.create_test_instances_and_names()
- BaseObject._has_implementation_of()

RNKuhns:
@fkiraly this should be ready for you to review. I been able to cover most of our base functionality outside of the TagAliaserMixin.

skbase/_base.py
```diff
@@ -427,7 +427,7 @@ def get_test_params(cls, parameter_set="default"):
         """
         # if non-default parameters are required, but none have been found, raise error
         if hasattr(cls, "_required_parameters"):
-            required_parameters = getattr(cls, "required_parameters", [])
+            required_parameters = getattr(cls, "_required_parameters", [])
```
fkiraly, line 430: I thought we removed the `_required_parameters` attribute?

At least `sktime` does not have it anymore, since required parameters are precisely thouse without a default in the init, and there are no kwargs.

skbase/_base.py
```diff
@@ -466,13 +466,8 @@ def create_test_instance(cls, parameter_set="default"):
         else:
             params = cls.get_test_params()
 
-        if isinstance(params, list):
```
fkiraly, line 469: is this covered by the test then?

What if a user implements sth but does not test?

skbase/tests/test_base.py
```diff
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
 # copyright: skbase developers, BSD-3-Clause License (see LICENSE file)
+# Elements of these tests re-use code developed in scikit-learn. These elements
```
fkiraly, line 3: should this also be added in sktime?

skbase/tests/test_base.py
```diff
@@ -170,6 +271,79 @@ def test_set_tags():
     assert FIXTURE_OBJECT_SET.get_tags() == FIXTURE_OBJECT_SET_TAGS, msg
 
 
+def test_set_tags_works_with_missing_tags_dynamic_attribute():
```
fkiraly, line 274: I would say, it should raise an error instead, because it means the object has not been correctly initialized!

fkiraly (changes requested):
Most of these make sense and are indeed useful additions.

My main comment (blocking until discussed, not necessarily actioned) is, many of these tests intersect with what `TestAllObjects` is doing, for concrete classes.

To avoid duplication, should we not instead run `check_object` with a test class?

As a test for `BaseObject`, using mock objects, rather than as a test for the object subjected to `check_object` - or is that too many inversions for your taste?

skbase/tests/test_base.py
```diff
@@ -251,26 +484,332 @@ def test_reset_composite():
...
+    new_base_obj = base_obj.clone()
+
+    assert base_obj.c == new_base_obj.c
+
+
+def test_clone_class_rather_than_instance_raises_error():
+    """Test clone raises expected error when cloning a class instead of an instance."""
+    msg = "You should provide an instance of scikit-learn estimator"
+    with pytest.raises(TypeError, match=msg):
+        clone(FIXTURE_EXAMPLE)
```
fkiraly, line 683: do we want to have an extraneous clone function?

skbase/tests/test_base.py
```diff
@@ -56,8 +92,73 @@ class FixtureClassChild(FixtureClassParent):
 FIXTURE_OBJECT_TAGS = {"A": 42424241, "B": 3, "C": 1234, 3: "E"}
 
 
+class InvalidInitSignatureTester(BaseObject):
+    def __init__(self, a, *args):
+        pass
+
+
+class Example(BaseObject):
```
fkiraly, line 100: to avoid side effects - since these are invoked by reference, and you could in theory write class variables in one test - would that be avoided by using proper pytest fixtures?

Seems complicated and low risk though.

skbase/tests/test_base.py
RNKuhns: > should this also be added in sktime?

Yes, I think having that probably makes sense as some of the sktime tests are also sklearn inspired.

skbase/tests/test_base.py
RNKuhns: Going the fixture route, shouldn't be too bad.

skbase/tests/test_base.py
RNKuhns: I think way back, when we added it to sktime, I had the logic setup to check for this and raise an error and you or another reviewer suggested that the code just initialize the missing attribute.

I don't have strong feelings either way, but this tests a feature of the current set_tags. I'd propose we leave the test, and if we decide to update how set_tags work, we amend the test (remove this test and replace with test that correct error is raised).

skbase/tests/test_base.py
RNKuhns: That's a good question. We are currently coupled with scikit-learn's `BaseEstimator` and separate clone function.

So for the time being, I think it makes sense to test the expected error is raised when interacting with the cloned functionality in scikit-learn. Eventually, we might remove the test (and the error raising it is related to) in the skbase `clone` method, because it is not a class method, and therefore, would be invoked on an instance by default.

skbase/_base.py
RNKuhns: This existed in the code at the time we ported it over. I think we should update the logic here and in other places to be in sync with sktime as part of this release. Ideally, we'd do that after we merge this PR so we have broader suite of tests in place.

RNKuhns:
> Another issue, `sktime` has some more functionality around fitted parameter inspection, although we can port that later.

I saw that, I think we want to have a PR that ports the additional `sktime` functionality as part of this release (e.g. we sync things back up).

RNKuhns:
> Most of these make sense and are indeed useful additions.
> 
> My main comment (blocking until discussed, not necessarily actioned) is, many of these tests intersect with what `TestAllObjects` is doing, for concrete classes.
> 
> To avoid duplication, should we not instead run `check_object` with a test class?
> 
> As a test for `BaseObject`, using mock objects, rather than as a test for the object subjected to `check_object` - or is that too many inversions for your taste?

I'm generally a fan of having separate tests for the base class and the concrete classes, but we can talk about it.

skbase/_base.py
RNKuhns: I just simplified the logic in the if-else statements.

skbase/tests/test_base.py
RNKuhns: Have now used fixtures.

skbase/tests/test_base.py
RNKuhns: For example, we might decide to have a clone function that clones duck-typed BaseObjects that comply with most of the interface, but don't have their own clone method.

RNKuhns:
> Looked in `sktime` whether what is there right now is all here, and found some diffs:
> 
> * `sktime` also has a test `test_components for the components inspection interface
> * `test_base_sktime` has some compatibility tests with `sklearn`

There is a `test_components` in `skbase`, but it was already there (not being added in this PR). The tests in `test_base_sktime` focus on the fitted parameter interface. We'll need to add those in the separate PR to add in the new functionality from sktime that has been added since we ported the original `BaseObject` into `skbase`.

skbase/_base.py
fkiraly: fair enough

skbase/tests/test_base.py
fkiraly: opened an issue here in `sktime`: https://github.com/sktime/sktime/issues/3853

skbase/tests/test_base.py
fkiraly: ok, makes sense as an interface compatibility test, with `scikit-learn`. That's a separate topic from whether we want to expose a separate `clone` ourselves.

skbase/_base.py
fkiraly: agreed on the point of keeping scope well-defined.

fkiraly (approved):
Agreed with resolutions and otherwise explanations where it turned out no action is needed in this PR.

There are some items that arise to do in other PR, mostly to do with updating current `BaseObject` with changes made since then in `sktime`, we should keep track of those: `_required_parameters` removed; functionality around fitting and fitted params.

PR #69, opened 2022-11-13
[BUG] fixes and improvements to `all_objects`

fkiraly:
This fixes some bugs and makes improvements to `all_objects` that are in `sktime`.

Bugs:

* subsetting should be via `issubclass`, not `isinstance`, since we filter classes, not objects. This seems to be correct in `sktime` but is wrong in `skbase`.

Improvements:

### DIFF
--- a/skbase/_lookup.py
+++ b/skbase/_lookup.py
@@ -774,12 +774,12 @@ def all_objects(
                 warnings.warn(str(e), ImportWarning)
 
     # Drop duplicates
-    all_estimators = list(set(all_estimators))
+    all_estimators = set(all_estimators)
 
     # Filter based on given estimator types
     def _is_in_object_types(estimator, object_types):
         return any(
-            inspect.isclass(x) and isinstance(estimator, x) for x in object_types
+            inspect.isclass(x) and issubclass(estimator, x) for x in object_types
         )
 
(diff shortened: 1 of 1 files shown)


fkiraly:
maybe we should also add a test for this?

RNKuhns:
@fkiraly there is still a bit of formatting weirdness in the ``all_object`` docstring when it renders in the read-the-docs check.

Do you mind taking a look at that and updating the format? I'm going to be opening up a separate PR for the tests.

fkiraly:
> Do you mind taking a look at that and updating the format? I'm going to be opening up a separate PR for the tests.

Just looked at it, looks ok to me?

I also did not change the general format or indentation, I think?

Do you mean the "note" boxes? I thought this was ok, but we can change the indentation.

RNKuhns:
@fkiraly I'm fine leaving the docstring as is for now. I thought I was seeing one part bolded that didn't seem like it should be (the boxes always through me for a loop, but I get its an indentation thing). I think we have an open issue related to the docstring for this, but I'll double-check when I add the tests for lookup functionality.

RNKuhns (approved):
LGTM. Will merge and handle tests and potential tweak to docstring in forthcoming PR.

PR #75, opened 2022-11-22
Update config files and remove use of setup.py

RNKuhns:

This updates the project config files and removes use of setup.py.


RNKuhns:
@fkiraly we need to update our CodeCov settings to provide access to the new repository (skbase). The repository CodeCov is still being routed to the old CodeCov. This assumes that it is going to a CodeCov repo for skbase. 

I don't see the ability to permission CodeCov to access skbase repo on my end. Let's take a look at this in our upcoming meeting.

fkiraly:
odd - here's the settings I can see in the org:
image

i.e., it looks like the codecov app has access?

RNKuhns:
@fkiraly the problem turned out to be with the need for "=" signs in the "addopts" configuration for pytest in the pyproject.toml versus other ways of specifying the configuration. This should be good to go, if it looks good to you.

fkiraly (approved):
Looks good to go, and thanks for the apostrophe.

Issue #32, opened 2022-07-28
design discussion: objects vs estimators


fkiraly:
Should estimators be in scope of the `baseobject` package?
If yes, what is the scope precisely?

I have been working on separating objects from estimators in `sktime`, see here: https://github.com/alan-turing-institute/sktime/pull/3122

RNKuhns:
We've included a BaseEstimator functionality with basic interface in ``skbase``. Closing this.

Issue #12, opened 2022-06-25
compatibility with sklearn `BaseEstimator`


fkiraly:
Question: how do we ensure compatibility with sklearn `BaseEstimator` and similar classes?

This might not be out of the box, if we check inheritance from `BaseObject` as opposed to availability of certain methods.

We could register a list of similar classes, but it would be better if this is out of the box compatible, e.g., `get_params` nesting and so on.

RNKuhns:
@fkiraly Whats the context for this? Are you thinking in terms of an eventual check interface (e.g. check_is_object) or something else? 

I'm in favor of duck typing that check but including a param to require a "strict" check (is also a sub class of BaseObject).

We might consider grouping the checks used in something like check_is_object by interface and having a parameter that controls what to check for (e.g. just get_params/set_params, tags, testing, etc). That would provide the most downstream comparability (sklearn baseobject could for instance could pass) and provide users the means to test for just the interface required in their use case.

In terms of the interface implementation: I believe the get_params/set_params interface uses duck typing, so it should work with like objects. One option is to ensure that remains and make sure the other interfaces (testing and tags) do the same.

RNKuhns:
Based on our call we also want to have tests that test the ability of estimators inheriting from `BaseObject` to pass scikit-learn's `check_is_estimator`.

To do this we need to look into sklearn's `check_is_estimator` to see the scope and format of sklearn's checks. If their checks are duck typed, a nice outcome is that estimators that inherit from `BaseObject` are able to pass sklearn's `check_is_estimator` directly. If the checks in sklearn's `check_is_estimator` are instead checking for direct inheritance from the sklearn `BaseEstimator` (in addition to checking the interface) we may need have tests that validate that you can construct an estimator object that inherits from sklearn's BaseEstimator and baseobject's `BaseObject` to get the `BaseObject` interface and pass sklearn's `check_is_estimator`.

fkiraly:
An alternative would be to check the main interoperation points: `get_params` and `set_params`.

That would require building composites either way round and see whether the nested access and write still works.

fkiraly:
From https://github.com/sktime/BaseObject/issues/17, another interoperation point is pretty printing.

fkiraly:
We should have a list of interoperation points we need to test, actually.

Issue #31, opened 2022-07-28
feature: contraction


fkiraly:
Not an immediate next step, but it would be great to have the feature of "contraction".

This is a function which takes a composite class and turns it into a non-composite class.
Parameters from the components are all mapped onto parameters of the result.

Example:
if input has `get_params().keys()` output `["component1", "component1_foo", "component2", "component2_bar", "foobar"]`, after contraction this should become `["foo", "bar", "foobar"]`, without any functional difference to the objects in question. In the "contracted" class, the components are hard-coded (whereas in the composite they may not have been).

ramcharanmedipally:
Hi there,
This is Avinash Kumar Dubey, I'm new on GitHub and want to contribute to OpenSource to make my Account robust.

I can add this contraction feature but I've some questions like where I should do the code for this and how this feature will work. for example,

`list = ["component1", "component1_foo", "component2", "component2_bar", "foobar"]`

method 1:
`contraction(list)`

method 2:
`list.contraction()`

fkiraly:
Hello @ramcharanmedipally!

Apologies, I did not see your post earlier since this repo is currently still in early design stage, and we didn't monitor this repo for new contributor posts.

This is not about list contraction but class contraction - apologies.

How about you look for interesting good first issues on `sktime`?
https://www.github.com/alan-turing-institute/sktime

Sorry again for not responding.

Abhishek9639:
Hi @fkiraly,
I’d like to work on this feature. I’ll start by understanding the expected behavior of class contraction and propose an implementation approach. Please assign this issue to me.
Thanks

PR #76, opened 2022-11-22
Initial `skbase` design proposal

### LINKED ISSUE #2: BaseObject design reconciliation points
There are some open points that we need to clarify with the @rnkuhns design. It also needs to work for `sktime`.

Some key points, from an architecture integration perspective:

* @rnkuhns merged the `BaseObject` and the `_HeterogeneousMetaEstimator`. I think this is not a good idea, it merges concerns, and overloads the `BaseObject` with pipeline functionality. I don't think `sktime` can live with that... too many concerns in one place, and refactor of all the pipelines would become a nightmare. I'd really prefer to move away from this god object direction...
* @rnkuhns removed the testing functions, but `sktime` needs those. These should also be public, not private, since other classes interact with them. So from a formal class design perspective, these need to be public unified interface points, even if they are not user facing.
* I like the addition of checking whether the tag containers exist, although instead of raising an error I simply would consider it empty
* `is_composite` is something I added out of a hurry. Perhaps we should sit down and design the components interface more cleanly?

RNKuhns:
#### Reference Issues/PRs

Fixes #2, #12, #32

This outlines `skbase`'s initial design expectations.

Provide initial feedback on the broad proposal (does the stated functionality make sense broadly, and if so, in the proposed sub-module layout). If that makes sense, we'll iterate on the more detailed design points until agreement is reached.


enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,172 @@
...
+Although the `BaseEstimator` interface may seem like it should include a `fit`
+method for learning the parameters from data, it is not included. Instead
+`BaseEstimator` assumes sub-classes implement a `fit` method (and that this
+method appropriately sets the `_is_fitted` attribute), since the signature of
+`fit` is learning task specific. Therefore, the specific `fit` implementation
+is left to child classes implemented outside of `skbase`.
+
+#### HeterogenousMetaObject
+
+**DESIGN DECISION: Do we want to provide specific base classes for useful
```
fkiraly, line 126: I think, yes, that would be useful!

Wonder whether the polymorphic base class would also be useful here: https://github.com/sktime/sktime/pull/3108

FYI @miraep8, @benheid re pipelines

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,172 @@
...
+
+The need to lookup classes arises in several contexts when working working with
+parametric objects, including the need to collect *similar* objects for
+testing or reporting.
+
+`skbase` provides this through two function interfaces:
+
+- `all_objects` provides the ability to recursively walk packages (or sub-packages)
+  and return the objects that meet the filters specified in the function call.
+- `get_package_metadata` provides the ability to recursively walk a package's moduels
```
fkiraly, line 152: nice idea to make this a public one

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,172 @@
...
+  objects and functions)
+
+#### `skbase.testing`: Testing BaseObjects
+TO BE ADDED.
+
+#### `skbase.validate`: Validating and Comparing BaseObjects
+
+### Example Repository
+This would be a simple example package that illustrates how `skbase`'s functionality
+can be used to create another package. It will also be used by us to test
```
fkiraly, line 163: hm, but it would not be the only way of testing it, right?

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,172 @@
...
+`skbase.testing`.
+
+### Template Repository
+To make it easy for others to use `skbase` in new projects, it is useful to provide
```
fkiraly, line 167: great idea!

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,172 @@
...
+  (per standard Python conventions).
+- Document all public attributes that are not parameters in the class docstring's
+  attributes section.
+
+#### BaseEstimator
+
+Scikit-learn style [estimators](https://scikit-learn.org/stable/tutorial/statistical_inference/settings.html?highlight=estimator#estimators-objects) are *"object that learn from data"*.
+
+In `scikit-learn` and `sktime` these can be *regressors*, *classifiers*,
+*clusterers*, *annotaters*, *transformers* and other type of classes implementing
```
fkiraly, line 103: typo: annotators

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,172 @@
...
+- `sktime` style interface for generating test instances
+- `scikit-learn` style interface for representing objects (e.g., pretty printing
+  and drawing a simple block representation in HTML)
+
+**DESIGN DECISION: Should we include a interface point for validating parameters?
+If we did this, we might have user-facing output that raises NotImplementedError
+if the users haven't defined non-public method that does the validation. This
+also might be something that falls just outside `skbase`.**
+
+`BaseObjects` should also follow certain design patterns and coding practices,
```
fkiraly, line 81: I would only code-typset class names, not declension letters - e.g., `BaseObject`-s

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,172 @@
...
+
+BaseObjects are base classes with:
+
+- `scikit-learn` style interface to get and set parameters
+- `sktime` style interface for working with *tags*
+- `sktime` style interface for cloning and re-instantiation (resetting)
+- `sktime` style interface for generating test instances
+- `scikit-learn` style interface for representing objects (e.g., pretty printing
+  and drawing a simple block representation in HTML)
+
```
fkiraly, line 75: composition should also be mentioned here

fkiraly (approved):
Excellent, reads well and makes a lot of sense design-wise!

Aware that there are a few open sections, but I don't mind what content goes in there (assuming it is a sensible extrapolate of the rest).

One possible set of features revolves around manipulating (base-)objects on the class level, e.g., composite class contraction: https://github.com/sktime/skbase/issues/31

I think this is pretty advanced though and may require working very close to `python` language features (dynamic generation of class definitions etc), so it may be the last of the things of the list to work on.

fkiraly:
(no change requests, since this is consolidating existing designs and ideas discussed on hackmd earlier, with some nice new ideas)

enhancement_proposals/sbep_1.md
RNKuhns: Nope, good point on that. But it would provide us some testing of a separate repository.

enhancement_proposals/sbep_1.md
RNKuhns: Fixed.

enhancement_proposals/sbep_1.md
RNKuhns: Good catch, I thought I'd fixed all those.

enhancement_proposals/sbep_1.md
RNKuhns: @fkiraly what are you thinking in terms of composition?

enhancement_proposals/sbep_1.md
RNKuhns: Great, I like the idea of making it easy for people to build pipelines. Since the goal of this design document is to provide the outline of the interface, we can defer the implementation details to a follow-up design specific to the class implementations.

@fkiraly, @miraep8 and @benHeid my initial thoughts are providing 3 meta classes: HeterogenousMetaObject (downstream users can make their own metaobjects), BasePipeline (makes it easier to make standard linear pipelines), BaseNetworkPipeline (makes it easier to build networked pipelines). But I'm up for a potential polymorphic approach to reduce the number of classes. For now we just need to state the use cases we intend to support. Do those  3 use cases make sense to everyone?

RNKuhns:
@fkiraly I've updated this PR to fill out the rest of the design proposal based on what we've talked about. Can you take another look to make sure I've represented everything correctly.

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,220 @@
...
+`skbase`'s primary API is provided through classes that allow for
+`scikit-learn`'s and `sktime`'s design patterns to be re-used in additional
+contexts. This includes:
+
+- [BaseObject](#BaseObject): a base class providing the package's primary
+  class level interface. Other classes are subclasses of `BaseObject`.
+- [BaseEstimator](#BaseEstimator): A subclass of `BaseObject` that adds a
+  high-level interface for *fittable* estimators
+- [BaseMetaObject](#BaseMetaObject): A subclass of `BaseObject`
```
fkiraly, line 61: hm, you're subclassing this. Is this fine?
What would an estimator be that is also a metaobject? Do we get a diamond inheritance diagram then?

enhancement_proposals/sbep_1.md
fkiraly: sorry did not see that. But, what I meant is what you wrote.

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,220 @@
...
+
+##### Base Pipeline Classes
+`skbase` intends to provide additional meta classes that expand on the `BaseMetaObject` API to include specific functionality
+for the type of common pipeline use cases found in `scikit-learn`, `sktime` and
+data processing workflows. This functionality will eventually be made available through two sub-classes of `BaseMetaObject`:
+
+- `BasePipeline` expands on the `BaseMetaObject` framework by providing
+  common functionality for (linear) stepped based pipelines like those found
+  in `scikit-learn` and `sktime`
+- `BaseDAGPipeline` expands on the `BaseMetaObject` framework by providing
```
fkiraly, line 151: well, this is probably going to be hairy, because it may require base class polymorphism.
Have you been involved with the pywatts discussion stream?

Although, in `skbase`, it might be easier since there is only one scitype (or two).

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,220 @@
...
+  functionality for  pipelines composed of directed-acyclic graphs (DAGs)
```
fkiraly, line 152: typo

fkiraly (approved):
+100.

Left some minor comments.

enhancement_proposals/sbep_1.md
fkiraly: yes, agreed. Sorry, did not see the reply earlier.

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,220 @@
...
+
+`skbase`'s core functionality will be available through submodule's tailored to
+a given use case.
+
+- `skbase.base` will include the `BaseObject` class and related base classes.
+- `skbase.lookup` will include the tools for retrieving (i.e., collectiong,
+  looking up) any `BaseObject`'s from a project.
+- `skbase.validate` will include tools for validating and comparing `BaseObject`'s
+  and/or collections of `BaseObject`'s.
+- `skbase.testing` will include tools for testing `BaseObject`s for interface
```
miraep8, line 45: 👍  I just wanted to say that I love the idea of skbase.validate and skbase.testing.  Especially think there is a space to also specify the tools as classes themselves. (eg a LengthChecker that can validate that certain specified inputs have the same type).

enhancement_proposals/sbep_1.md
miraep8: Question - do we need a BasePipeline class at all? Or do we simply need to make a sufficiently flexible pipeline class?  I think having a BasePipeline class opens one up to making many types of pipeline classes (which is kind of what we do in sktime now tbh).  But a better way to do this is perhaps to just build a pipeline that can handle sufficient complexity of different use cases.

RNKuhns: @miraep8 I think that is a valid point for a single package. But the end-stage goal is for `skbase` to be used external to `sktime` (I for instance have other use-cases and the hope is it would make creating a new package easy for someone else too). In that case, a pipeline in a different package would be able to inherit some base functionality, but the implementation of the "action" or "state-changing" methods that the pipeline performs would be up to the external developer.

fkiraly:
@RNKuhns, fine with me - feel free to merge when you are happy with this being in its final form.

enhancement_proposals/sbep_1.md
fkiraly: agreed! Did not notice it, but I agree with @miraep8 that tests as objects would be nice.

enhancement_proposals/sbep_1.md
```diff
@@ -0,0 +1,220 @@
...
+  collections of `BaseObject`s.
+- [Base pipeline classes](#Base-Pipeline-Classes): Subclasses of `BaseMetaObject`
+  that provide generic functionalty for common pipeline use cases.
+
+#### BaseObject
+
+BaseObjects are base classes with:
+
+- `scikit-learn` style interface to get and set parameters
+- `sktime` style interface for working with *tags*
```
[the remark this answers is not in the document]
fkiraly: here's an old, related, STEP that never got implemented: https://github.com/sktime/enhancement-proposals/blob/main/steps/05_scitype_based_IO_checks/step.md

I would do it differently today, but just FYI

fkiraly:
What do we do with this PR, @RNKuhns, @miraep8?

RNKuhns:
> What do we do with this PR, @RNKuhns, @miraep8?

@fkiraly I plan to finish this by early next week.

Want to look at things 1 more time.

enhancement_proposals/sbep_1.md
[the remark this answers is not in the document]
RNKuhns: @miraep8 from the `skbase` perspective we want to make it possible to use "tags". We leave the usage to the individual packages (but I agree there can be non-tag ways to accopmlish the same outcome. But I do believe there are cases where tags can make the most sense).

enhancement_proposals/sbep_1.md
RNKuhns: @fkiraly aren't forecasting pipelines really estimators that are also metaobjects? That being said, I went the Mixin route in the final version.

enhancement_proposals/sbep_1.md
RNKuhns: It is an interesting idea, though beyond the intended scope of this proposal. I do think we should revisit this topic when we dive further into the `skbase.testing` interface (hopefully in the not too distant future).

Going this route, testing is really executing a pipeline of "checker" or "validator" objects (which makes intuitive sense). This could also be useful in parameter validation.

RNKuhns:
> > What do we do with this PR, @RNKuhns, @miraep8?
> 
> @fkiraly I plan to finish this by early next week.
> 
> Want to look at things 1 more time.

This is now ready to merge. Will do so an include in the v0.3.0 release.

PR #78, opened 2022-11-22
Refactor sub modules

RNKuhns:
#### Reference Issues/PRs

Partial implementation of skbase design proposal.

Refactors `skbase` functionality into `skbase.base`, `skbase.lookup`, `skbase.validate`. Presence of `skbase.testing` remains unchanged.


skbase/validate/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+"""Tools for validating and comparing BaseObjects and collections of BaseObjects."""
+import inspect
+from collections.abc import Iterable
+from typing import List
+
+__author__: List[str] = ["RNKuhns", "fkiraly"]
+__all__: List[str] = []
+
+
+def _check_list_of_str_or_error(arg_to_check, arg_name):
```
fkiraly, line 13: I thought it was best practice to not have any functionality in `__init__` (except related to the concern of generating exports)

RNKuhns:
@fkiraly note that I've removed the comment from lgtm-com because we are using CodeQL now (that is name of lgtm functionality that was ported into Github). We need to disable lgtm-com app integration.

skbase/validate/__init__.py
RNKuhns: Good catch, I was moving things around quickly. I've fixed this.

RNKuhns:
@fkiraly I believe this has refactored everything into the different modules as noted in the design proposal in #76. Should be ready for your review.

fkiraly (approved):
Makes much sense.

I suppose this is the equivalent of sorting your garden gnome collection into different cabinets and then arrange them in regular geometric patterns. We can all relate.

Some non-blocking comments:

* should `_remove_single` also be in `_nested_seq`? It unnests.
* `_nested_seq`, perhaps better `_nested_iter` or `_nested_list` (knowing it also applies to tuples)?

RNKuhns:
> Makes much sense.
> 
> I suppose this is the `sktime` equivalent of sorting your garden gnome collection into different cabinets and then arranging them in regular geometric patterns. We can all relate.

I like that analogy! And yep, that's the plan, get everything in an order so it is easier to add the rest of the functionality from here.

PR #79, opened 2022-11-23
Fixes to issues identified by CodeQL Scanning

RNKuhns:

 This includes 2 fixes to errors that were raised by #68. Fixing this will allow #68 to be merged.

Fixes call init on parent (BaseObject) in test classes. Also, updates conf.py to fix minor issues with unused variables.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -16,7 +16,6 @@ import sys
 
 import skbase
 
-# sys.path.insert(0, os.path.abspath("."))
 # -- Path setup --------------------------------------------------------------
 
 # When we build the docs on readthedocs, we build the package and want to
@@ -37,7 +36,7 @@ author = "skbase Developers"
 
 
 # The full version, including alpha/beta/rc tags
-# release = skbase.__version__
+release = skbase.__version__
 
 # -- General configuration ---------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -65,9 +66,9 @@
     extensions += ["sphinx_sitemap"]
 
     # -- Sitemap -------------------------------------------------------------
-    html_baseurl = os.environ.get("SITEMAP_URL_BASE", "http://127.0.0.1:8000/")
```
fkiraly, line 68: what was this??

skbase/tests/test_base.py
```diff
@@ -177,6 +177,7 @@ def __init__(self, foo, bar=84):
         self.foo = foo
         self.foo_ = deepcopy(foo)
         self.bar = bar
+        super().__init__()
```
fkiraly, line 180: I think we should test for that this is caught.

In `sktime` it is caught by checking whether `_tags_dynamic` is initialized.

fkiraly (approved):
looks ok, assuming the mysterious URL related changes are unimportant.

Perhaps you can remove the commented code too?

docs/source/conf.py
RNKuhns: @fkiraly it got copied from elsewhere.

skbase/tests/test_base.py
RNKuhns: I think that makes sense. Will verify whether we have the test and if not add it as part of #62.

RNKuhns:
> looks ok, assuming the mysterious URL related changes are unimportant.
> 
> Perhaps you can remove the commented code too?

Commented out code removed, since that part wasn't needed.

PR #80, opened 2022-11-23
Add unit tests of lookup functionality and minor lookup refactor

### LINKED ISSUE #35: [BUG] `all_objects` docstring renders strangely
List of parameters renders as a single block

### LINKED ISSUE #66: Expand Test Coverage of Lookup Interface
We need to expand test coverage of `skbase`'s lookup interface.

Additional tests of the following functionality:

- `_is_ignored_module`
   - `_filter_by_class`
   - `_filter_by_tags`
   - `_walk`
   - `_import_module`
   - `all_objects`
     - Test ignore_modules is not None
     - Test when module name starts with leading underscore
     - Test with suppress_import_stdout == False
     - Test return_names == True
     - Test return_tags == True
     - Test as_dataframe == True
     - Test object_types filtering functionalty
     - Test exclude_estimators functionality
     - Test filter tags functionality
   - `get_package_metadata`

### LINKED ISSUE #70: rename `ignore_modules` to `modules_to_ignore`?
In the `all_objects` module, we currently have inconsistent naming of the same type of argument, `ignore_modules` and `modules_to_ignore`.

I think we should consistently rename this to `modules_to_ignore`, because `ignore_modules` sounds like a boolean condition (e.g., "ignore modules?" - yes/no).

RNKuhns:
#### Reference Issues/PRs

Fixes #70, improves test coverage (fixes #66), fixes bugs, improves code consistency and docstrings (fixes #35 ).

This adds unit tests for the lookup functionality in `skbase`. Also does minor refactor and docstring improvements as things are discovered in testing process.


fkiraly:
What exactly is stuck here, do we know that? Can I help?

RNKuhns:
> What exactly is stuck here, do we know that? Can I help?

@fkiraly I haven't investigated the POSIX paths not having starts with attribute in detail yet. Those aren't occurring locally.

Others are just things popping up as I am slowly adding tests. Expect to finish up the test adding today or tomorrow.

RNKuhns:
> > What exactly is stuck here, do we know that? Can I help?
> 
> @fkiraly I haven't investigated the POSIX paths not having starts with attribute in detail yet. Those aren't occurring locally. I'm checking now if the issue is with the validation of the returned results in the test or call to the actual _walk function having some sort of bug. Will let you know.

There was a minor issue in the _walk function. Its documentation said it accepted a Path. But the underlying pkgutil.iter_modules only accepts strings. Function was updated to convert pathlike objects to strings prior to passing them to pkgutil.iter_modules.

RNKuhns:
@fkiraly I believe I've got tests covering everything that is "reasonable" to test (there are a few edge cases for raising import errors and a few other things, that are not easy to replicate in the repository and won't be tested), except for `sktime` compatability checks.

Note that I'm running into an error when using `skbase.lookup.all_objects` on the `sktime` package. Specifically, "Skipped: Skipping MLflow tests for Windows and macOS". We'll need to investigate that (may be we are catching certain errors, but not the one raised in this specific instance) and add the compatability check.

RNKuhns:
> Note that I'm running into an error when using `skbase.lookup.all_objects` on the `sktime` package. Specifically, "Skipped: Skipping MLflow tests for Windows and macOS". We'll need to investigate that (may be we are catching certain errors, but not the one raised in this specific instance) and add the compatability check.

Figured the error out, just needed the correct values for `modules_to_ignore`.  Also, `skbase.lookup.all_objects` is filtering objects that inherit from `skbase.base.BaseObject` so we'll need to do the initial compatability check in `sktime` PR. So I think this is in good shape to review.

RNKuhns:
@fkiraly this is still ready for your review. Just pushed a deprecation of a parameter name in `get_package_metadata` (replaced with alternative name) and added some doc readability/clarity improvements.

fkiraly:
>  Specifically, "Skipped: Skipping MLflow tests for Windows and macOS". 

Probably coming from these lines in `test_mlflow_sktime_model_export`:

```python
if not sys.platform.startswith("linux"):
    pytest.skip("Skipping MLflow tests for Windows and macOS", allow_module_level=True)
```

Not sure why this causes an exception though.

Maybe moving this to a `pytest.mark` would solve the issue?

fkiraly:
> Just pushed a deprecation of a parameter name

I don't think we need deprecation as long as we don't have users :-)
(that will of course change soon 😁 )

pyproject.toml
```diff
@@ -87,6 +87,7 @@ test = [
     "safety",
     "numpy",
     "scipy",
+    "pandas",
```
fkiraly, line 90: ah, for the optional `pandas` output, I see.
As long as we don't make it a hard dep, it's fine, I suppose.

skbase/lookup/_lookup.py
```diff
@@ -110,7 +109,13 @@ def _is_non_public_module(module_name: str) -> bool:
     is_non_public : bool
         Whether the module is non-public or not.
     """
-    is_non_public: bool = "._" in module_name
+    if not isinstance(module_name, str):
+        raise ValueError(
+            f"Parameter `module_name` should be str, but found {type(module_name)}."
+        )
+    is_non_public: bool = "._" in module_name or (
+        "." not in module_name and module_name.startswith("_")
```
fkiraly, line 117: should this not simply be
`"._" in module_name or module_name.startswith("_")`?

Because the current condition does not cover a module without submodule starting with `_`.

skbase/lookup/_lookup.py
```diff
@@ -194,28 +203,36 @@ def _filter_by_tags(
     """
     if tag_filter is None:
         return True
+
     if hasattr(klass, "get_class_tags"):
         klass_tags = klass.get_class_tags()
+
+    else:
+        # If the object doesn't have tag interface it can't have the tags in tag_filter
+        return False
```
fkiraly, line 212: interesting question: the user expected behaviour may differ in this case, depending on whether a default or a non-default is being filtered by.

For simplicity, I agree with this, but I wonder whether there is a nicer solution. Not a blocking comment.

pyproject.toml
fkiraly: Not a blocking comment.

skbase/lookup/_lookup.py
```diff
@@ -233,7 +250,7 @@ def _walk(root, exclude=None, prefix=""):
 
     Parameters
     ----------
-    root : Path
+    root : Path-like
```
fkiraly, line 253: should we say "i.e., `Path` or `Path` coercible"?

skbase/lookup/_lookup.py
```diff
@@ -257,42 +276,44 @@ def _walk(root, exclude=None, prefix=""):
...
     suppress_import_stdout: bool = True,
-    loader: importlib.machinery.SourceFileLoader = None,
 ) -> ModuleType:
     """Import a module, while optionally suppressing import standard out.
 
     Parameters
     ----------
-    module_name : str
-        Name of the module to be imported.
+    module : str or importlib.machinery.SourceFileLoader
```
fkiraly, line 286: were the old `module` and `loader` args always mutually exclusive?

fkiraly: looking through the old code, I suppose so - makes sense to unify the argument therefore

skbase/lookup/_lookup.py
```diff
@@ -475,11 +509,13 @@ def get_package_metadata(
     path: Optional[str] = None,
     recursive: bool = True,
     exclude_non_public_items: bool = True,
+    exclude_non_public_modules: bool = True,
```
fkiraly, line 512: duplicate argument in two different spellings?

fkiraly: ah, I see - deprecation.
I don't think we need deprecation at this stage (no users or dependencies), but as you prefer.

Unless, of course you are a user and need deprecation.

skbase/lookup/_lookup.py
```diff
@@ -556,20 +613,36 @@ def get_package_metadata(
         - "contains_base_objects": whether any module classes that
           inherit from ``BaseObject``.
     """
+    # If someone changed deprecated parameter from the default we'll use it but warn
+    if exclude_nonpublic_modules is not True:
```
fkiraly, line 617: better maybe: `if not exclude_nonpublic_modules`?

skbase/lookup/_lookup.py
```diff
@@ -624,127 +707,202 @@ def get_package_metadata(
 def all_objects(
     object_types=None,
     filter_tags=None,
+    exclude_objects=None,
     exclude_estimators=None,
```
fkiraly, line 711: same as above - do we need deprec?

skbase/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+    "FixtureClassChild",
+    "CompositionDummy",
+    "ResetTester",
+    "InvalidInitSignatureTester",
+    "RequiredParam",
+    "Buggy",
+    "ModifyParam",
+    "NoParamInterface",
+]
+__author__: List[str] = ["RNKuhns"]
```
fkiraly, line 19: I think some of these are lifted from my code in `sktime.base.test_base`?

skbase/lookup/tests/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
+"""Tests of skbase lookup functionality."""
```
fkiraly, line 2: do we need `sktime` BSD license blurb in some places here?

skbase/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+    """Class for testing invalid signature."""
+
+    def __init__(self, a, *args):
+        super().__init__()
+
+
+class RequiredParam(BaseObject):
+    """BaseObject class with _required_parameters."""
+
+    _required_parameters = ["a"]
```
fkiraly, line 95: Did we want to remove the `_required_parameters` interface?
As in, required parameters being exactly those in `__init__` without a default?

skbase/tests/conftest.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def __init__self(self, a, b=7):
```
fkiraly, line 97: typo? Why did nothing pick this up?

skbase/mock_package/mock_package.py
```diff
@@ -37,3 +37,22 @@ def get_test_params(cls, parameter_set="default"):
         params1 = {"foo": 42}
         params2 = {"foo": CompositionDummy(126)}
         return [params1, params2]
+
+
+class InheritsFromBaseObject(BaseObject):
+    """A class inheritting from BaseObject."""
```
fkiraly, line 43: should be "inheriting", no?

skbase/lookup/tests/test_lookup.py
```diff
@@ -0,0 +1,899 @@
...
+    "is_concreate_implementation",
+    "is_base_class",
+    "is_base_object",
+    "authors",
+    "module_name",
+]
+REQUIRED_FUNCTION_METADATA_KEYS = ["func", "name", "description", "module_name"]
+
+
+@pytest.fixture
```
fkiraly, line 71: do we need these fixtures if classes are returned?

Fixtures are typically used for python objects that have identity, but classes are not among those (objects would be)

skbase/lookup/tests/test_lookup.py
```diff
@@ -0,0 +1,899 @@
...
+    """Fixture to root path of skbase package."""
+    return fixture_test_lookup_mod_path.parent
+
+
+def _check_package_metadata_result(results):
+    """Check output of get_package_metadata is expected type."""
+    result_okay: bool = True
+    if not (isinstance(results, dict) and all(isinstance(k, str) for k in results)):
+        result_okay = False
+    for k in results:
```
fkiraly, line 124: more pythonic? `for k, mod_metadata in results.items():`

skbase/lookup/tests/test_lookup.py
```diff
@@ -0,0 +1,899 @@
...
+
+
+@pytest.fixture
+def fixture_skbase_root_path(fixture_test_lookup_mod_path):
+    """Fixture to root path of skbase package."""
+    return fixture_test_lookup_mod_path.parent
+
+
+def _check_package_metadata_result(results):
+    """Check output of get_package_metadata is expected type."""
```
fkiraly, line 120: this function is hard to read.

* do you want to turn all instances of `result_ok =False; break (then stuff happens)` into a simple `return False`?
* the `if`/`elif` is also a bit confusing, as it implies a "switch" - could simply be a lot of "`if ... return False`"?

fkiraly (changes requested):
Great!

* tests make sense overall. Impressed that we tests internal utilities as well as the full function.
* tests seem to cover important arguments of `all_objects`
* left some minor/non-blocking comments overall

My main blocking comment or question:
* overall, I don't see a test that compares retrieved objects with a hard-coded set of expected objects from the mock package, e.g., in `test_all_objects_returns_expected_types`. Maybe I'm just blind/confused
* `test_all_object_class_lookup`, should this not test the content of the return, `objs`, being the right objects (given the mock package)?

pyproject.toml
RNKuhns: Yep, these aren't installed when the package is installed. Just in the CI and optionally by users, so this should let us handle other soft dependencies this way too (though we'd like to minimize them too).

skbase/lookup/_lookup.py
RNKuhns: @fkiraly, that's right, it is simpler and will accomplish the same thing.

skbase/lookup/_lookup.py
RNKuhns: @fkiraly what do you mean by default or non-default?

I did waffle in this, but felt this was safer of two simple options (other being classes without tag interface are not filtered at all).

skbase/lookup/_lookup.py
RNKuhns: I don't feel strongly about this. I like trying to keep type description of param short if possible. But Path or path coercive works.

skbase/lookup/_lookup.py
RNKuhns: I am a user, but don't need deprecation!

I'm fine removing the deprecation here.

skbase/lookup/_lookup.py
RNKuhns: Yep, more pythonic. Will push commit.

skbase/lookup/_lookup.py
RNKuhns: This is more of a question for you and sktime.


If you are planning to wrap skbase's function in a sktime one,  you could handle deprecation there.

My thoughts are let's not deprecate the new functions added in skbase (e.g., get_package_metadata). I'm fine not deprecating the old ones too, but had been due to potential difference relative to sktime.

skbase/tests/conftest.py
RNKuhns: I believe they are. You moved them to skbase.tests.test_base. I moved them out of there to make them available to other tests in other modules more naturally (in conftest.py).

skbase/tests/conftest.py
RNKuhns: It is still hanging about until we sink with updated sktime functionality. Given our desire for release 0.3.0 soon. I think it is release 0.4.0 issue (say mid-January)

skbase/tests/conftest.py
RNKuhns: I'll verify, but I believe it is only being used in test designed to fail. Let me make sure that an init typo wasn't what we are checking!

skbase/lookup/_lookup.py
fkiraly: ok, makes sense, especially since it would otherwise pick up all (python) objects.

skbase/lookup/_lookup.py
fkiraly: well, either is a bit vague - maybe something more specific, or examples?

skbase/lookup/_lookup.py
fkiraly: I don't need deprecation either, happy with the direct change.

skbase/lookup/_lookup.py
fkiraly: happy without deprecation - I know what's coming, after all, and there is no hard dep which could break it for users

skbase/tests/conftest.py
fkiraly: well, so I should be author, is what I mean

skbase/tests/conftest.py
fkiraly: ah, makes sense.

skbase/tests/conftest.py
fkiraly: makes sense

skbase/lookup/_lookup.py
RNKuhns: I went with: str path or path-like. Then in the description portion said it can be a string path, pathlib.Path or other path-like object.

skbase/lookup/_lookup.py
RNKuhns: Yea, it was redundant to have the two params. It was just one param, but doing different things based on type.

skbase/lookup/_lookup.py
RNKuhns: I've added a commit to remove the deprecation. That keeps life easy for now.

skbase/lookup/_lookup.py
RNKuhns: Also removed this deprecation.

skbase/tests/conftest.py
RNKuhns: Yep, good catch. I'll get that updated.

skbase/mock_package/mock_package.py
RNKuhns: Fixed the typo.

skbase/lookup/tests/__init__.py
RNKuhns: Added it to test_lookup.py, which is where the tests that are affected live.

skbase/tests/conftest.py
RNKuhns: Typo wasn't the intent. But the class was used in tests designed to fail, so we didn't see it. Fixed now.

skbase/lookup/tests/test_lookup.py
RNKuhns: Agreed. I've made the update.

skbase/lookup/tests/test_lookup.py
RNKuhns: I've made the suggested changes. Agreed it isn't super easy to read, hopefully that helps.

skbase/lookup/tests/test_lookup.py
RNKuhns: Yep, that makes sense. I'll reserve them for things like objects, dicts and the like.

RNKuhns:
> Great!
> 
> * tests make sense overall. Impressed that we tests internal utilities as well as the full function.
> * tests seem to cover important arguments of `all_objects`
> * left some minor/non-blocking comments overall
> 
> My main blocking comment or question:
> 
> * overall, I don't see a test that compares retrieved objects with a hard-coded set of expected objects from the mock package, e.g., in `test_all_objects_returns_expected_types`. Maybe I'm just blind/confused
> * `test_all_object_class_lookup`, should this not test the content of the return, `objs`, being the right objects (given the mock package)?

That's a good point. I'd been avoiding a specific comparison and instead verifying things based on whether the return types, counts, etc make sense in light of the filters being applied on the returned metadata/objects. I'll push a test of the specific output as well. I'll make that separate from the `test_all_objects_returns_expected_types` and `test_all_object_class_lookup`, which are, respectively, focused on whether all_objects is returning the right type of results (not that the exact results are correct and the usage of class_lookup param in all_objects.

fkiraly (approved):
Looks good to me now.

Checked:
* reference in `TestAllObjects`
* mock package looks good
* lookup also uses mock package

All comments addressed, hence.

Issue #83, opened 2022-11-29
[ENH] Port updated base class functionality from sktime


RNKuhns:
**Is your feature request related to a problem?**

Since `skbase` initially ported over functionality from `sktime` there has been parallel development. The features should be synced up. 

**Describe the solution you'd like**

Sync up features in `skbase` with those in `sktime`.

RNKuhns:
While under the hood working on this, it probably makes sense to also fix #65 and #81

fkiraly:
sure!

fkiraly:
should I do this?

RNKuhns:
> should I do this?

It's yours!

RNKuhns:
@fkiraly do we know of any other functionality that hasn't been ported over?

fkiraly:
only this, I think:
https://github.com/sktime/skbase/issues/124

fkiraly:
done

PR #86, opened 2022-12-28
[ENH] Equality dunder for `BaseObject`-s

fkiraly:
Mirrors `sktime` PR https://github.com/sktime/sktime/pull/3862, partial towards #83

Description from `sktime` PR:

This adds an equality dunder to `BaseObject` descendants.

Two `BaseObject` descendants are considered equal if they define the same estimator blueprint, i.e., nested `get_params` result in value-identical dictionaries (with nested call to `__eq__` should those dictionaries contain `BaseObject` descendants).

Includes a test for expected properties of the dunder.

This is prompted by issues of equality testing in tests such as in the failure relevant for https://github.com/sktime/sktime/pull/3857, where equality currently means object identity rather than parameter identity.

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -77,6 +77,24 @@ class BaseObject(_BaseEstimator):
         self._tags_dynamic = {}
         super(BaseObject, self).__init__()
 
+    def __eq__(self, other):
+        """Equality dunder. Checks equal class and parameters.
+
+        Returns True iff result of get_params(deep=False)
+        results in equal parameter sets.
+
+        Nested BaseObject descendants from get_params are compared via __eq__ as well.
+        """
+        from skbase.testing.utils.deep_equals import deep_equals
+
+        if not isinstance(other, BaseObject):
+            return False
+
+        self_params = self.get_params(deep=False)
+        other_params = other.get_params(deep=False)
+
+        return deep_equals(self_params, other_params)
+
     def reset(self):
         """Reset the object to a clean post-init state.
 
--- a/skbase/testing/utils/deep_equals.py
+++ b/skbase/testing/utils/deep_equals.py
@@ -9,8 +9,6 @@ Objects compared can have one of the following valid types:
 from inspect import isclass
 from typing import List
 
-import numpy as np
-
 from skbase.testing.utils._dependencies import _check_soft_dependencies
 
 __author__: List[str] = ["fkiraly"]
@@ -79,7 +77,9 @@ def deep_equals(x, y, return_msg=False):
         if res is not None:
             return _pandas_equals(x, y, return_msg=return_msg)
 
-    if isinstance(x, np.ndarray):
+    if _is_npndarray(x):
+        import numpy as np
+
         if x.dtype != y.dtype:
             return ret(False, f".dtype, x.dtype = {x.dtype} != y.dtype = {y.dtype}")
         return ret(np.array_equal(x, y, equal_nan=True), ".values")
@@ -88,12 +88,6 @@ def deep_equals(x, y, return_msg=False):
         return ret(*_tuple_equals(x, y, return_msg=True))
     elif isinstance(x, dict):
         return ret(*_dict_equals(x, y, return_msg=True))
-    # np.nan returns False when comparing np.nan == np.nan
-    # we deal with this by looking at the type
-    elif isinstance(x, type(np.nan)):
-        return ret(
-            isinstance(y, type(np.nan)), f"type(x)={type(x)} != type(y)={type(y)}"
-        )
     elif isclass(x):
         return ret(x == y, f".class, x={x.__name__} != y={y.__name__}")
     elif type(x).__name__ == "ForecastingHorizon":
@@ -118,6 +112,13 @@ def _is_pandas(x):
         return True
     else:
         return False
+
+
+def _is_npndarray(x):
+
+    clstr = type(x).__name__
+    if clstr in ["ndarray"]:
+        return True
 
 
 def _pandas_equals(x, y, return_msg=False):
--- a/skbase/testing/utils/deep_equals.py
+++ b/skbase/testing/utils/deep_equals.py
@@ -6,6 +6,7 @@ Objects compared can have one of the following valid types:
     pd.Series, pd.DataFrame, np.ndarray
     lists, tuples, or dicts of a valid type (recursive)
 """
+from inspect import isclass
 from typing import List
 
 import numpy as np
@@ -43,6 +44,7 @@ def deep_equals(x, y, return_msg=False):
         indication of what is the reason for not being equal
             concatenation of the following strings:
             .type - type is not equal
+            .class - both objects are classes but not equal
             .len - length is not equal
             .value - value is not equal
             .keys - if dict, keys of dict are not equal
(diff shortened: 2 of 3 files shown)


skbase/base/_base.py
```diff
@@ -77,6 +77,24 @@ def __init__(self):
...
 
+    def __eq__(self, other):
+        """Equality dunder. Checks equal class and parameters.
+
+        Returns True iff result of get_params(deep=False)
+        results in equal parameter sets.
+
+        Nested BaseObject descendants from get_params are compared via __eq__ as well.
+        """
+        from skbase.testing.utils.deep_equals import deep_equals
```
RNKuhns, line 88: Is there a use-case where we'd want this logic outside of a BaseObject?

It may make sense to add a public function to skbase.validate that performs this comparison and then import it here.

RNKuhns: Note that we could also open up an Issue to do that later.

skbase/testing/utils/deep_equals.py
```diff
@@ -70,14 +70,21 @@ def ret(is_equal, msg):
     # we now know all types are the same
     # so now we compare values
 
+    # flag variables for available soft dependencies
+    pandas_available = _check_soft_dependencies("pandas", severity="none")
+    numpy_available = _check_soft_dependencies("numpy", severity="none")
+
+    if numpy_available:
```
RNKuhns, line 77: Should we factor numpy logic into helper function like in pandas? Answer could be that the logic is simpler, so it is fine to include directly.

skbase/testing/utils/deep_equals.py
```diff
@@ -105,6 +119,23 @@ def _is_pandas(x):
         return False
 
 
+def _is_npndarray(x):
```
RNKuhns, line 122: Logic looks a bit weird here. Shouldn't we be checking that the name of the class includes "ndarray". 

Also agree with the CodeQL check. It seems like good practice to inlude the else False condition (even though None is returned by default and it won't be interpretted as True).

skbase/testing/utils/deep_equals.py
```diff
@@ -105,6 +119,23 @@ def _is_pandas(x):
...
+    """Coerce x to list."""
+    if not isinstance(x, (list, tuple)):
+        x = [x]
+    if isinstance(x, tuple):
+        x = list(x)
+
+    return x
+
+
 def _pandas_equals(x, y, return_msg=False):
```
RNKuhns, line 139: Doesn't pandas have testing functions to assert objects are equal? Is there a reason we aren't leveraging those?

skbase/tests/test_base.py
```diff
@@ -1036,3 +1039,60 @@ def test_has_implementation_of(
...
+    assert non_composite_2 != non_composite_3
+    assert composite == composite_2
+    assert composite != composite_3
+    assert composite_2 != composite_3
+
+    # equality should not be affected by fitting
+    composite.fit()
+    non_composite_2.fit()
+
+    assert non_composite == non_composite
```
RNKuhns, line 1091: Let's add a test that uses dunder comparison after the clone/copy functionality. Should result in equality, but good to check interaction between the different functionalities.

skbase/tests/test_base.py
```diff
@@ -1036,3 +1039,60 @@ def test_has_implementation_of(
...
+    assert composite == composite
```
RNKuhns, line 1092: For these tests in general, can we reduce the number of asserts by taking advantage of the fact that if a == b, then we only need to test a == c and not b == c?  These are also really fast tests, so there isn't harm in leaving them.

RNKuhns (changes requested):
I've got some minor change requests.

Note that some of my comments are questions. But I think we should consider moving the logic in __eq__ to its own function in skbase.validate. And we should add a case in the dunder equality test that tests this with our clone/copy functionality (expect a clone to equal what it was cloned from).

skbase/base/_base.py
fkiraly: well, `deep_equals` is typically used in tests.
In `sktime`, it is not imported outside tests except here.

skbase/testing/utils/deep_equals.py
fkiraly: well, I was considering this, but I don't know how we would deal with lines 105ff, i.e., where `!=` results in an iterable, and if `numpy` is available, you want to use `np.all` (due to efficiency), otherwise `python` `all`.

skbase/testing/utils/deep_equals.py
fkiraly: yes, you are right - not sure what I was thinking. Replaced the logic with the "obvious" check

skbase/testing/utils/deep_equals.py
fkiraly: not added in this PR, but originally added to allow for dealing with nested `pandas`. `pandas` onboard checkers don't (or didn't) play along well with the case where you nest `pandas` in `pandas`, since `!=` on `pandas` results in not a boolean but an iterable.

The case where you nest `pandas` in `pandas` occurs for the `nested_univ` mtype, the "original" mtype used for panel tasks in `sktime` (but now discouraged since `pandas` is acting up on newer versions when it comes to nesting itself).

skbase/tests/test_base.py
fkiraly: done

skbase/tests/test_base.py
fkiraly: I think: no, and the reason is subtle.
I too saw the code quality comment.

The point is that we *do not already know* that equality as implemented is transitive, but *we need to test it* because we just implemented an equality that *should* satisfy that property.

I.e., theoretically, yes, it should be that a==c and a==b implies b==c - but we need to test whether actuality matches theory (rather than assume it).

skbase/testing/utils/deep_equals.py
fkiraly: generally, `deep_equals` supports equality checking where you arbitrarily nest the supported types, including `pandas`. Given the funny composite types in time series, I think it is fair to have it in `sktime`. In `skbase`, don't know.

skbase/base/_base.py
fkiraly: not sure what the best thing to do is. Recommendations?

RNKuhns: Lets leave as is and open an Issue to consider porting the logic to `skbase.validate` too.

skbase/testing/utils/deep_equals.py
RNKuhns: Ah, yea, lets leave as is then. We know it works this way and its limited upside.

skbase/tests/test_base.py
RNKuhns: good point. If in first assert a==b, and in second assert a==c (or a!=c) we know the result for b==c (or b!=c). Likewise, if the first assert fails and a!=b, then when a==c (or a!=c), we know the outcome of b==c  if a==c, but if a!=c we don't know if b will (b/c it isn't equal to a).

skbase/testing/utils/deep_equals.py
RNKuhns: Yea -- I considered asking the same question about the _fh comparisons. In the end we can't be all things to all packages. And it may not make sense to have time series specific items, but for now we can leave them. It would be good if `sktime` deep_equals could build on the `skbase` one (so we could remove the time series specific parts from `skbase`). Lets try to keep that in mind in the `sktime` PR that uses the `skbase` functionality.

fkiraly:
> And we should add a case in the dunder equality test that tests this with our clone/copy functionality (expect a clone to equal what it was cloned from).

Added those as test cases.

RNKuhns (approved):
LGTM now. We'll open the issue about whether to have separate function to test class equality.

PR #87, opened 2022-12-28
[ENH] `get_fitted_params` interface for `BaseEstimator`

fkiraly:
Uniform `get_fitted_params` interface for `BaseEstimator`, including tests.
Partially addresses #83.

Related `sktime` issues and PR:
* https://github.com/sktime/sktime/issues/971
* https://github.com/sktime/sktime/pull/3077
* https://github.com/sktime/sktime/issues/3381
* https://github.com/sktime/sktime/pull/3598
* https://github.com/sktime/sktime/issues/3645

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -814,6 +814,11 @@ class BaseEstimator(BaseObject):
     Extends BaseObject to include basic functionality for fittable estimators.
     """
 
+    # tuple of non-BaseObject classes that count as nested objects
+    # get_fitted_params will retrieve parameters from these, too
+    # override in descendant class - common choice: BaseEstimator from sklearn
+    GET_FITTED_PARAMS_NESTING = ()
+
     def __init__(self):
         """Construct BaseEstimator."""
         self._is_fitted = False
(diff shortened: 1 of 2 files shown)


skbase/base/_base.py
```diff
@@ -851,3 +856,101 @@ def check_is_fitted(self):
...
+        fitted_params = {}
+
+        def sh(x):
+            """Shorthand to remove all underscores at end of a string."""
+            if x.endswith("_"):
+                return sh(x[:-1])
+            else:
+                return x
+
+        # add all nested parameters from components that are sktime BaseObject
```
RNKuhns, line 886: Should be objects that are skbase BaseEstimator's

skbase/base/_base.py
```diff
@@ -851,3 +856,101 @@ def check_is_fitted(self):
...
+                c_f_params = comp.get_fitted_params()
+                c_f_params = {f"{sh(c)}__{k}": v for k, v in c_f_params.items()}
+                fitted_params.update(c_f_params)
+
+        # add non-nested fitted params of self
+        fitted_params.update(self._get_fitted_params())
+
+        # add all nested parameters from components that are sklearn estimators
+        # we do this recursively as we have to reach into nested sklearn estimators
+        n_new_params = 42
```
RNKuhns, line 899: Why is this 42 to start? Won't be obvious to future developers, so we need a comment to explain.

skbase/base/_base.py
```diff
@@ -851,3 +856,101 @@ def check_is_fitted(self):
...
+        Returns
+        -------
+        fitted_params : dict with str keys
+            fitted parameters, keyed by names of fitted parameter
+        """
+        obj = obj if obj else self
+
+        # default retrieves all self attributes ending in "_"
+        # and returns them with keys that have the "_" removed
+        fitted_params = [attr for attr in dir(obj) if attr.endswith("_")]
```
RNKuhns, line 936: Can these dictionary comprehensions not be combined with a compound if statement?

skbase/base/_base.py
```diff
@@ -814,6 +814,11 @@ class BaseEstimator(BaseObject):
     Extends BaseObject to include basic functionality for fittable estimators.
     """
 
+    # tuple of non-BaseObject classes that count as nested objects
+    # get_fitted_params will retrieve parameters from these, too
+    # override in descendant class - common choice: BaseEstimator from sklearn
+    GET_FITTED_PARAMS_NESTING = ()
```
RNKuhns, line 820: Why is this not a tag?

skbase/tests/test_baseestimator.py
```diff
@@ -81,3 +82,45 @@ def test_check_is_fitted_raises_error_when_unfitted(fixture_estimator_instance):
...
+        self.bar = bar
+
+    def fit(self):
+        """Fit, dummy."""
+        if hasattr(self.foo_, "fit"):
+            self.foo_.fit()
+        self._is_fitted = True
+
+
+def test_get_fitted_params():
```
RNKuhns, line 102: Should the test class and test also look for a simple  fitted param? E.g., in ffit something like some_fitted_param_ gets a value assigned?

RNKuhns (changes requested):

skbase/base/_base.py
fkiraly: it's more a config for the entire package than a tag that is estimator specific.
Not sure what the best way is - I changed it from being `sklearn` `BaseEstimator` (hard-coded) in `sktime`.

skbase/base/_base.py
fkiraly: I thought about that but didn't think the alternatives I came up with were easier to read. Suggestions always appreciated.

skbase/base/_base.py
fkiraly: fixed

skbase/base/_base.py
fkiraly: renamed variables and added explanation. Should be much clearer now, hopefully.

skbase/tests/test_baseestimator.py
fkiraly: That's what is being tested in line 121, `assert set(non_comp_f_params.keys()) == {"foo"}`

Or, do you mean sth different? If so, kindly expand.

skbase/base/_base.py
RNKuhns: Aren't we using tags as things set by developers, and configs more things set by end-users? This would seem like something that is set once by the developer using `skbase`, instead of an end-user. But I do agree, it is more of a package level thing. But maybe end users would want to turn it off (not sure, but who knows)?

Let's keep this in mind when we work on the interface for package level config.

skbase/tests/test_baseestimator.py
RNKuhns: Ah -- I didn't see where `foo` was being assigned. It is assigned to `foo_` attribute in `__init__ ` and not `fit` so I missed it. Seems like your doing that for ease to test composite and non-composite case. It works, and proves that an attribute that ends in "_" but doesn't start with "_" will get returned as a fitted param.

skbase/base/_base.py
```diff
@@ -851,3 +856,103 @@ def check_is_fitted(self):
...
+                c_f_params = comp.get_fitted_params()
+                c_f_params = {f"{sh(c)}__{k}": v for k, v in c_f_params.items()}
+                fitted_params.update(c_f_params)
+
+        # add non-nested fitted params of self
+        fitted_params.update(self._get_fitted_params())
+
+        # add all nested parameters from components that are sklearn estimators
+        # we do this recursively as we have to reach into nested sklearn estimators
+        any_components_left_to_process = True
```
RNKuhns, line 899: This is much eaiser to read/follow. Thanks for updating the variable names and adding comments.

RNKuhns (approved):
Looks good to me now. Up to you whether you want to update the list comprehensions. But that is not required (ultimately, the extra time to traverse the lists multiple times should be negligible given that most classes will have "relatively" few attributes).

skbase/base/_base.py
fkiraly: two more lines due to linting, but I suppose it is easier to read as it puts pairs of conditions together that "match"

fkiraly: changed

skbase/base/_base.py
fkiraly: yes, you are right!
It is a developer thing, and it might be worth setting at the estimator level.

I now start agreeing witih you that it probably should be a tag.

The one question I still have then is, there are currently no tags in `skbase`. This would open a can of worms of "tags already defined by `skbase` (although the trade-off is to explaining that attribute).

skbase/tests/test_baseestimator.py
fkiraly: so, this is resolved then, right?

PR #95, opened 2022-12-31
[ENH] in `QuickTester.run_tests`, rename `return_exceptions` argument to `raise_exceptions`

fkiraly:
This PR changes the name of the `return_exceptions` argument in `QuickTester.run_tests` to `raise_exceptions`. Parallels https://github.com/sktime/sktime/pull/4030, see there for reasoning why.

This PR is without deprecation (since we can still avoid it).

### DIFF
--- a/skbase/testing/test_all_objects.py
+++ b/skbase/testing/test_all_objects.py
@@ -233,7 +233,7 @@ class QuickTester:
     def run_tests(
         self,
         obj,
-        return_exceptions=True,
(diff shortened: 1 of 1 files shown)


RNKuhns (approved):
Looks straight-forward and I like raise_exceptions naming better (and I see it fits with logic, since return_exceptions was used to calculate raise_exceptions internally).

Will merge once CI runs on updated main.

PR #96, opened 2023-01-02
[ENH] remove `sktime` references from dependency checkers

fkiraly:
This PR removes `sktime` references from docstrings and error messages in the dependency checkers, and streamlines the message generation logic a bit.


RNKuhns (approved):
Looks reasonable to me. Straight-forward and appears to accomplish its stated purpose.

PR #99, opened 2023-01-04
Remove deprecation

RNKuhns:
This removes deprecation added in prior PR refactoring the package to `skbase.base`, `skbase.looup`, `skbase.validate` and `skbase.testing` modules.

The package will use deprecation in the future (likely within the next few releases), but this early in the development is not going to.


fkiraly (approved):
Looks reasonable. I trust you caught all of it.

In `sktime`, the convention is to add the full version number directly next to any planned change, with a description of the action, so a repo search gets you to everywhere where a change is needed.

PR #105, opened 2023-01-12
Create Issue and PR Templates

RNKuhns:
This adds issue and PR templates for the repository.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
...
+---
+
+**Describe the bug**
+<!--
+A high-level description of what the bug is. Make sure to clearly and concisely describe where the bug was encountered, what happened and what you expected to happen. Provide more details on each of these in the sections below.
+-->
+
+**To Reproduce**
+<!--
+Provide the information necessary to reproduce the bug. If the bug occurred when executing ``skbase`` related code, provide a reproducible code example. If the bug arose in another context, provide instructions for recreating the bug.
```
fkiraly, line 17: maybe worth saying: optimally, the code should not rely on external data sources, and should be minimally reproducible, perhaps also a link to sth like this
https://stackoverflow.com/help/minimal-reproducible-example

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Expected behavior**
+<!--
+Provide more detail on what you expected to happen.
+-->
+
+
+**Environment**
+<!--
+Please provide information about the environment you encountered the bug in. If your bug was related to running ``skbase`` Python code, provide details about your Python environment (e.g., operating system and version, version of Python, version of ``skbase``, version of other related packages, etc).
```
fkiraly, line 29: is there an easier way, i.e., more instruction based, less language based, to ensure people provide the right information here? E.g., sth like please run `print_versions` or similar

.github/ISSUE_TEMPLATE/maintenance-issue.md
```diff
@@ -0,0 +1,15 @@
+---
+name: Maintenance issue
```
fkiraly, line 2: should maintenance and CI/CDe be separate? I think they tend to be highly intersectional. Since both issue types are also rare, why not just merge then?

.github/ISSUE_TEMPLATE/bug_report.md
RNKuhns: That is what we want. Makes sense to be more direct indicating that.

.github/ISSUE_TEMPLATE/maintenance-issue.md
RNKuhns: I had gone back and forth about it. I'm fine combining them.

.github/ISSUE_TEMPLATE/bug_report.md
RNKuhns: I'm going to leave this for later.

PR #108, opened 2023-01-13
Update pydocstyle in pre-commit config

RNKuhns:
#### Reference Issues/PRs
There was a commit in `pydocstyle` v6.2.3 that has since been reverted. Version 6.2.4 has not yet been released, so this reverts back to 6.2.2.

Also updating additional dependency section to make use of extra dependency installation. Note that `pydocstyle` switched to using tomli/tomllib for toml parsing instead of toml. Their "extra" install maps this automatically.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,7 +58,7 @@ repos:
     rev: 6.2.3
     hooks:
       - id: pydocstyle
-        additional_dependencies: ["toml"]
+        additional_dependencies: ["tomli"]
 
   - repo: https://github.com/nbQA-dev/nbQA
     rev: 1.6.1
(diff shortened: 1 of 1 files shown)


fkiraly:
> There was a commit in `pydocstyle` v6.2.3 that has since been reverted. Version 6.2.4 has not yet been released, so this reverts back to 6.2.2.


Sorry, I don't get this, could you kindly explain? Is the commit in `pydocstyle` itself, or in this or another repo, in relation to `pydocstyle`?

RNKuhns:
@fkiraly the issue is in pydocstyle itself. They already have a fix, but it is not released yet. So this is using prior patch version until the new one is out.

While reading their documentation, I also saw their change with toml parsing and figured I'd update to follow their suggested approach for installing their optional dependency.

fkiraly (approved):
Ah, so some bug with 6.2.3. Fine with me then.

Issue #110, opened 2023-01-13
[ENH] `from skbase import BaseObject` should work!


fkiraly:
I think that `from skbase import BaseObject` should work like it used to.
Same for key objects like `BaseEstimator`, `all_objects`, etc.

Why: as it is currently designed, typical users will mostly import `skbase` for `BaseObject` and perhaps a small one-digit number of functions or classes.

I missed this in the PR that restructured the package - I feel strongly about this but not too strongly.

PR #113, opened 2023-01-14
[ENH] export key objects at root

fkiraly:
This PR adds exports at `skbase` root level, for `BaseObject`, `BaseMetaObject`, `BaseEstimator`, and `all_objects`.

Fixes https://github.com/sktime/skbase/issues/110

Of course it's worth discussing whether we want this, first (in #110)

### DIFF
--- a/skbase/__init__.py
+++ b/skbase/__init__.py
@@ -8,7 +8,10 @@ sktime design principles in your project.
 """
 from typing import List
 
+from skbase.base import BaseEstimator, BaseMetaEstimator, BaseObject
+from skbase.lookup import all_objects
+
 __version__: str = "0.3.0"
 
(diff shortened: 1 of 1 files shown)


fkiraly:
Arrrghhhhh same test failure as per https://github.com/sktime/skbase/issues/114
and this time I don't even know where I would have to add the expected function in the config...

RNKuhns:
@fkiraly I personally am not a fan of having things exportable from multiple public facing modules (e.g., skbase.base or skbase.lookup and also skbase). To me it creates ambiguity in expected usage by our (eventual) downstream users.

fkiraly:
alright then. Let's keep it as is if you feel that way - no too strong opinions.

PR #118, opened 2023-01-15
[ENH] `skbase` native `get_params`

### LINKED ISSUE #116: [ENH] `get_params` is missing
It would appear that `get_params` is missing from `skbase` - it is obtained from `sklearn` but should instead be part of `skbase`.
(given that we want to transfer all joint base class functionality from `sktime` and `sklearn` into `skbase`)

fkiraly:
Adds an `skbase` native `get_params` to `BaseObject` which replaces the `sklearn` based one. Fixes #116.

The new `get_params` is largely based on the `sklearn` one, with minor changes (better docstring, dependency on `skbase` methods).
As far as I can see, this removes the last logical dependency on `sklearn` `BaseEstimator`.
Credit should already be given through the blanket reference to `sklearn` `BaseEstimator`.

Pretty printing and pretty display still constitute `sklearn` dependencies of `BaseObject`.

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -74,6 +74,7 @@ class BaseObject(_BaseEstimator):
     """
 
     def __init__(self):
+        """Construct BaseObject."""
         self._tags_dynamic = {}
         super(BaseObject, self).__init__()
 
@@ -704,6 +705,7 @@ class TagAliaserMixin:
     deprecate_dict = {"old_tag": "0.12.0", "tag_to_remove": "99.99.99"}
 
     def __init__(self):
+        """Construct TagAliaserMixin."""
         super(TagAliaserMixin, self).__init__()
 
     @classmethod
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -244,7 +244,7 @@ class BaseObject(_BaseEstimator):
         params = {key: getattr(self, key) for key in self.get_param_names()}
 
         if deep:
-            for key, value in params:
+            for key, value in params.items():
                 if hasattr(value, "get_params"):
                     deep_items = value.get_params().items()
                     params.update({f"{key}__{k}": val} for k, val in deep_items)
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -249,8 +249,8 @@ class BaseObject(_BaseEstimator):
                 if hasattr(value, "get_params"):
                     deep_items = value.get_params().items()
                     deep_params.update({f"{key}__{k}": val for k, val in deep_items})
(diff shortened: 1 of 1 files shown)


skbase/base/_base.py
```diff
@@ -216,6 +217,42 @@ def get_param_defaults(cls):
...
+    def get_params(self, deep=True):
+        """Get a dict of parameters values for this object.
+
+        Parameters
+        ----------
+        deep : bool, default=True
+            If True, will return a dict of parameter name : value for this object,
+            including parameters of components (= BaseObject-valued parameters).
+            If False, will return a dict of parameter name : value for this object,
+            but not include parameters components.
```
RNKuhns, line 229: I think it should be "but not include paremeters of components" or  "but not include component parameters".

skbase/base/_base.py
```diff
@@ -216,6 +217,42 @@ def get_param_defaults(cls):
...
+
+        Returns
+        -------
+        params : dict with str-valued keys
+            keys-value pairs include:
+            * always: all parameters of this object, as via `get_param_names`
```
RNKuhns, line 235: I think you need a blank line before after the text you want to be bulleted. Otherwise, it won't render correctly.

RNKuhns: Since there is no text after the part you want to be bulleted, in this case I think you want to have a blank line between the "key-value pairs include:" and the first bullet.

RNKuhns (changes requested):
@fkiraly requestd changes are minor. Just want to make sure the docstring renders correctly before this is merged.

skbase/base/_base.py
fkiraly: ah, of course. Or `parameters' components` which is confusing, so I changed it to your first suggestion.

skbase/base/_base.py
fkiraly: done

fkiraly:
resolved both change requests, see above

fkiraly:
augh, those pre-commit failures are now.
I assume you have them all fixed in here? https://github.com/sktime/skbase/pull/119

RNKuhns:
> augh, those pre-commit failures are new. I assume you have them all fixed in here? #120

Yep -- #120 will clear this up. @fkiraly , can you take a quick look and approve? Once that's approved, I'll go ahead and start getting some of these other PRs unblocked.

RNKuhns (approved):
LGTM now.

fkiraly:
Sure, opened a PR for `numpy` here: https://github.com/sktime/skbase/pull/121

PR #121, opened 2023-01-24
[MNT] numpy as a soft dependency

fkiraly:
This PR adds `numpy` as a soft dependency, see discussion in https://github.com/sktime/skbase/pull/111.

It should be noted that this will have an effect only once the `sklearn` dependency is removed or moved to be a soft dependency, as `numpy` is a core dependency of `sklearn`.


RNKuhns (approved):
LGTM. Thanks for adding.

PR #122, opened 2023-01-26
[ENH] Initial type and named object validator code

RNKuhns:
##### For all contributions

##### For code contributions


fkiraly (changes requested):
This looks very useful!

One comment, based on a learning in `sktime` and a guess on what this will be for:

I recently changed the checking logic to allow lists of tuples of arbitrary length, where element 0 is name/str, element 1 is estimator, and the remaining elements can be anything (e.g., index specifiers in column estimators etc).

This is because there were a number of examples where heterogeneous estimators had more than two elements, and it turned out to be easiest to relax the typing constraint.

RNKuhns:
@fkiraly I've got updates and additional testing taken care of.

I'm thinking we initially restrict the named object format to the current (str, BaseObject) or Dict[str, BaseObject], then relax it in a follow-on PR. The reason I want to take a bit more time to get familiar with different use cases of named objects and consider if there are any specific patterns we want to check. For example, (str, BaseObject, str). And whether we want to allow certain dictionary formats for these options (maybe Dict[str, (BaseObject, str)] but that is just off the top of my head).

fkiraly (approved):
ok, sure - happy to do this in a separate PR, I will open an issue.

Looks fine otherwise.

Issue #49, opened 2022-09-30
[ENH] raise test coverage to >80%


fkiraly:
We need to raise test coverage of the package to higher, this will likely include:

* tests for the testing framework
* porting of base class tests from sktime
* writing new base class tests

Running List of Needed Tests:

- #63 
- #61
- #86
- #66

RNKuhns:
@fkiraly this makes sense. I'm going to update the issue to include a list of all the different functionality we aren't currently testing. Then we can make individual issues for the items in the list.

RNKuhns:
#62 should handle #61 and #63. I'll move onto the `skbase._lookup` functionality while you review.

RNKuhns:
We've met this threshold. Closing.

Issue #19, opened 2022-07-01
[DOC] Add examples to docstrings


RNKuhns:
#### Describe the issue linked to the documentation

Improve the initial package documentation by adding examples to the docstrings.

RNKuhns:
This was completed for the `skbase.validate` module. Will work to incorporate it in other modules prior to release v0.5.0.

PR #126, opened 2023-02-05
[ENH] Improve `skbase.utils`

RNKuhns:
#### Reference Issues/PRs

Completes the portion of #49 and #19 for the `skbase.utils` submodule. Note that I've included docstring examples for the functions even if they aren't public because I think it is helpful for developers to be able to get a quick glimpse of what the function does.

Also updates names of functions to be non-public (I'm guessing we don't want them in user API) and make minor tweaks to make functions more robust or generalizable (as discovered during test creation).

The changes reflect:

- Adding basic tests of `skbase.utils` (functions are tested further in tests of their downstream uses)
- Improved docstrings of `skbase.utils` functionality (I believe these all now comply with `numpydoc` standard and have examples)
- Renamed functions to be non-public and updated the names in import statements and calls elsewhere in the package
- made minor improvements to functionality based on things I noticed when creating the tests/documenting the functionality that would easily make it more robust or improve its generality.

No new dependencies.

#### What should the reviewer focus their review on
I believe I added reasonable tests of the basic functionality of these utils (they are implicitly further tested in their downstream uses). So the key thing is whether you agree with the renaming to make the functionality non-public (i.e., not part of public API). Let me know if any of the functions in `skbase.utils._iter.py` or `skbase.utils._nested_iter.py` seem like something we want to include in the public API. Happy to switch the naming convention and add them to the API Reference.

##### For all contributions

##### For code contributions

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -77,3 +77,21 @@ Testing
     BaseFixtureGenerator
     QuickTester
     TestAllObjects
+
+Utils
+=====
+
+.. automodule:: skbase.utils
+    :no-members:
+    :no-inherited-members:
+
+.. currentmodule:: skbase.utils
+
+.. autosummary::
+    :toctree: api_reference/auto_generated/
+    :template: function.rst
+
+    flatten
+    is_flat
+    unflat_len
+    unflatten
--- a/skbase/base/_meta.py
+++ b/skbase/base/_meta.py
@@ -11,7 +11,7 @@ from inspect import isclass
 from typing import List
 
 from skbase.base._base import BaseEstimator
-from skbase.utils._nested_iter import _flatten, _is_flat, _unflatten
+from skbase.utils._nested_iter import flatten, is_flat, unflatten
 
 __author__: List[str] = ["mloning", "fkiraly"]
 __all__: List[str] = ["BaseMetaEstimator"]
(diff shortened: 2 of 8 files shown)


RNKuhns:
@fkiraly do you want to take a look at this PR. Just cleans up some of the utility functionality.

fkiraly (changes requested):
Apologies, I thought I had posted here already.

I am very happy with the improved docstrings, tests, etc.

However, I think at least some of the functions (e.g., `flatten`, `unflatten`) should remain public - in `sktime`, these are used in some concrete ensembles where they are useful to deal with index selection.

I do not have a too strong feeling about it though and would not block the PR.
However, leaving them private would mean that `sktime` has to copy-paste and maintain its own utilities around nested iterables, which seems wrong to me.

RNKuhns:
@fkiraly I am definitely up for leaving some functions part of the public API. It's an easy update on this PR.

Which do you think we should specifically make part of the public API?

RNKuhns:
@fkiraly let me know what you think in terms of the functions in utils you want public. I can switch the naming conventions and add them to the API really quick.

fkiraly:
> Note that I've included docstring examples for the functions even if they aren't public because I think it is helpful for developers to be able to get a quick glimpse of what the function does.

I won't complain :-)

> @fkiraly I am definitely up for leaving some functions part of the public API. It's an easy update on this PR.
Which do you think we should specifically make part of the public API?

`flatten`, `unflatten`. Maybe `is_flat`.

fkiraly (changes requested):
Looks good (no strong opinion about flatten/unflatten, I would prefer them public)

Question, before we merge: there is an intersection with PR #130, the file `_iter`. Is this intended? If not, should be fixed; if so, is there a sequence this is intended to be merged?

RNKuhns:
@fkiraly I'm down to make those three public. As you pointed out There is some interaction with #130. I'll merge that first and push the changes for this one and we can merge it tomorrow.

RNKuhns:
@fkiraly this is failing because of fix needed for `pre-commit.ci` failure unrelated to the changes (that's in #137). Can you take a look at the changes and see if they meet expectations. If they look good, you can approve on condition of tests passing once #137 is merged and I'll merge things in order tomorrow.

fkiraly (approved):
Happy with this - the public/private was not a blocker, I just wanted to flag the intersection. Given that you have a plan on how to handle the intersection/dependency here, all is good from my side.

PR #136, opened 2023-02-23
Add object_type param to named object check

### LINKED ISSUE #135: [ENH] Add ability to specify specific class types to named object checks

It would be useful for `is_named_objects` and `check_named_objects` to be able to narrow validations to only allow specific types of BaseObjects.

It would be useful for `is_named_objects` and `check_named_objects` to accept a parameter of the specific type(s) of named objects (e.g., class types) that are allowed for the checks.

RNKuhns:
#### Reference Issues/PRs

Fixes #135

Allows users to specify a specific class type to be used in the named object checks (`is_sequence_named_objects` and `check_sequence_named_objects`. 

No

##### For all contributions

##### For code contributions

### DIFF
--- a/skbase/validate/_named_objects.py
+++ b/skbase/validate/_named_objects.py
@@ -33,7 +33,7 @@ def is_sequence_named_objects(
     seq_to_check: Union[Sequence[Tuple[str, BaseObject]], Dict[str, BaseObject]],
     allow_dict: bool = True,
     require_unique_names=False,
-    object_type: Optional[type] = None,
+    object_type: Optional[Union[type, Tuple[type]]] = None,
 ) -> bool:
     """Indicate if input is a sequence of named BaseObject instances.
 
@@ -69,8 +69,8 @@ def is_sequence_named_objects(
           depends on whether `seq_to_check` follows sequence of named
           BaseObject format.
 
-    object_type : class, default=None
-        The class type that is used to ensure that all elements of named objects
+    object_type : class or tuple[class], default=None
+        The class type(s) that is used to ensure that all elements of named objects
         match the expected type.
 
     Returns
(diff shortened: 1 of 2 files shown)


RNKuhns:
@fkiraly this should be good to go besides the fix in #137 that solves the `pre-commit.ci` failure.

skbase/validate/_named_objects.py
```diff
@@ -68,6 +69,10 @@ def is_sequence_named_objects(
           depends on whether `seq_to_check` follows sequence of named
           BaseObject format.
 
+    object_type : class, default=None
```
fkiraly, line 72: I would say that it can also be a tuple of class, this might be useful, e.g., in pipelines

skbase/validate/_named_objects.py
```diff
@@ -157,6 +174,7 @@ def check_sequence_named_objects(
     seq_to_check: Union[Sequence[Tuple[str, BaseObject]], Dict[str, BaseObject]],
```
fkiraly, line 174: I really don't like this additional amount of non-functional code generated from `overload`, but it's a matter of taste and not a blocker for me...

skbase/validate/_named_objects.py
```diff
@@ -222,6 +243,9 @@ def check_sequence_named_objects(
         - If False, then whether or not the function returns True or False
           depends on whether `seq_to_check` follows sequence of named BaseObject format.
 
+    object_type : class, default=None
```
fkiraly, line 246: same here

fkiraly (approved):
Happy to merge.

I would suggest to extend the contract to allow tuples of classes, which it already does anyway (but doesn't say so in the docstring, and doesn't check in the tests).
Why: pipelines can be heterogeneous, and we may like to cover that case.

RNKuhns:
Good point on tuples. Will update docstring and tests pre-merge.

PR #138, opened 2023-03-03
[ENH] tag manager mixin

fkiraly:
port of https://github.com/sktime/sktime/pull/3630 to `skbase`

This PR factors out the tag management logic into a `_FlagManager` mixin, and generalizes the functionality so it can be used for multiple fields of tags.

Advantages:
* increases cohesion and reduces coupling, separates concerns in the `BaseObject` class: state/parameter management and tag management
* prepares the ground for managing configuration fields, e.g., for implementing a `set_config` similar to the new `sklearn` feature

Test coverage is already there for the base case, but should be added for a different tag field. Will do that with the PR that adds config handling.

(I called the flag manager since not all flags are tags, and interactions with the sklearn `_get_tags`)

A STEP is available here: https://github.com/sktime/enhancement-proposals/pull/29

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -76,7 +76,7 @@ class BaseObject(_FlagManager, _BaseEstimator):
 
     def __init__(self):
         """Construct BaseObject."""
-        self._tags_dynamic = {}
+        self._init_flags(flag_attr_name="_tags")
         super(BaseObject, self).__init__()
 
     def __eq__(self, other):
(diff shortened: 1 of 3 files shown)


RNKuhns (changes requested):
@fkiraly see the requested docstring requests on #140 (which appears to have some logic repeated there -- likely since you needed some of this logic to add config interface).

fkiraly:
Addressed, see replies in #140

RNKuhns (approved):
LGTM now.

PR #139, opened 2023-03-04
[ENH] sync `TestAllObjects` with `sktime`

fkiraly:
This syncs `TestAllObjects` with the current `sktime` state:

* added docstrings
* some descriptive error messages
* sharper logic in testing the output of `clone` in `test_clone`

### DIFF
--- a/skbase/testing/test_all_objects.py
+++ b/skbase/testing/test_all_objects.py
@@ -527,9 +527,10 @@ class TestAllObjects(BaseFixtureGenerator, QuickTester):
         Also tests inheritance and super call logic in the constructor.
 
         Tests that:
+
         * create_test_instance results in an instance of estimator_class
-        * __init__ calls super.__init__
-        * _tags_dynamic attribute for tag inspection is present after construction
+        * `__init__` calls `super.__init__`
+        * `_tags_dynamic` attribute for tag inspection is present after construction
         """
         object_instance = object_class.create_test_instance()
 
@@ -716,6 +717,7 @@ class TestAllObjects(BaseFixtureGenerator, QuickTester):
         """Check that clone method does not raise exceptions and results in a clone.
 
         A clone of an object x is an object that:
+
         * has same class and parameters as x
         * is not identical with x
         * is unfitted (even if x was fitted)
@@ -738,6 +740,7 @@ class TestAllObjects(BaseFixtureGenerator, QuickTester):
         Assumes test_create_test_instance has passed and certified create_test_instance.
 
         Tests that:
+
         * constructor has no varargs
         * tests that constructor constructs an instance of the class
         * tests that all parameters are set in init to an attribute of the same name
--- a/skbase/testing/test_all_objects.py
+++ b/skbase/testing/test_all_objects.py
@@ -518,7 +518,19 @@ class TestAllObjects(BaseFixtureGenerator, QuickTester):
     """Package level tests for BaseObjects."""
 
     def test_create_test_instance(self, object_class):
-        """Check first that create_test_instance logic works."""
+        """Check create_test_instance logic and basic constructor functionality.
+
+        create_test_instance and create_test_instances_and_names are the
+        key methods used to create test instances in testing.
+        If this test does not pass, validity of the other tests cannot be guaranteed.
+
+        Also tests inheritance and super call logic in the constructor.
+
+        Tests that:
+        * create_test_instance results in an instance of estimator_class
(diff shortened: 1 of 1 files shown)


skbase/testing/test_all_objects.py
```diff
@@ -518,7 +518,19 @@ class TestAllObjects(BaseFixtureGenerator, QuickTester):
...
+        """Check create_test_instance logic and basic constructor functionality.
+
+        create_test_instance and create_test_instances_and_names are the
+        key methods used to create test instances in testing.
+        If this test does not pass, validity of the other tests cannot be guaranteed.
+
+        Also tests inheritance and super call logic in the constructor.
+
+        Tests that:
+        * create_test_instance results in an instance of estimator_class
```
RNKuhns, line 530: I don't think this is going to render correctly in sphinx. I think that bulleted lists need a space between the list bullets and the text above or below

RNKuhns: So in this case you'd need something like:


```python
"""Some doc summary.

Some extended summary.

Tests that:

* first bullet
* second bullet
* third bullet
"""
````

File: skbase/testing/test_all_objects.py [hunk 2]
```diff
@@ -694,18 +713,42 @@ def test_set_params_sklearn(self, object_class):
             assert is_equal, msg
 
     def test_clone(self, object_instance):
-        """Check we can call clone from scikit-learn."""
-        object_instance.clone()
-        # object_clone = object_instance.clone()
-        # assert deep_equals(object_clone.get_params(), object_instance.get_params())
+        """Check that clone method does not raise exceptions and results in a clone.
+
+        A clone of an object x is an object that:
```
RNKuhns, line 718: same as above.

skbase/testing/test_all_objects.py
```diff
@@ -694,18 +713,42 @@ def test_set_params_sklearn(self, object_class):
...
-        assert getfullargspec(object_class.__init__).varkw is None
+        """Check that the constructor has sklearn compatible signature and behaviour.
+
+        Based on sklearn check_estimator testing of __init__ logic.
+        Uses create_test_instance to create an instance.
+        Assumes test_create_test_instance has passed and certified create_test_instance.
+
+        Tests that:
+        * constructor has no varargs
+        * tests that constructor constructs an instance of the class
```
RNKuhns, line 742: Same as above.

RNKuhns (changes requested):
@fkiraly some quick docstring suggestions.

For some reason our read-the-docs integration that builds the docs as part of PRs doesn't seem to be working. I'll try to get that back up and running.

Might be worth building the docs locally and taking a look at how these render.

skbase/testing/test_all_objects.py
fkiraly: resolved in all instances

skbase/testing/test_all_objects.py
fkiraly: same, resolved

skbase/testing/test_all_objects.py
fkiraly: same, resolved

fkiraly:
formatting stuff is resolved as far as I can see.

RNKuhns:
> > For some reason our read-the-docs integration that builds the docs as part of PRs doesn't seem to be working. I'll try to get that back up and running.
> 
> I don't quite get how readthedocs integration works to modify contents of the PR. I thought there was an automatic doc build, and readthedocs then gets the latest version from the `main` branch? But readthedocs doesn't actually trigger the automatic doc build.
> 
> I might be mistaken, so I would appreciate some explanation to understand what you mean.


It doesn't modify the PR. But we do have ReadTheDocs setup to build the docs for each PR. So it is the PR that triggers RTD to build the docs as one of the checks we run on PRs. I know I'm a stickler for documentation -- but I've been using it to make sure things are rendering like we'd want. 

For some reason the RTD build of the PR docs wasn't working for about a day. But it is working now. So I'll take a look and wrap this up.

RNKuhns:
@fkiraly the lists are rendering better. There are a few extra things that aren't rendering exactly like we'd want. But I won't hold this up for that now. We'll take another pass across the `skbase.testing` documentation ahead of upcoming presentation.

RNKuhns (approved):
I'm approving, since we plan to take another pass at the testing module documentation ahead of presentation.

We'll catch the remaining minor docstring formatting items at that time.

PR #140, opened 2023-03-04
[ENH] object config interface

fkiraly:
port of https://github.com/sktime/sktime/pull/3822 to `skbase`

This PR adds a config management capabilities to all `sktime` objects, similar to `sklearn`'s new config interface, via two methods `get_config` and `set_config`.

It is based on the flag manager mixin introduced in https://github.com/sktime/sktime/pull/3630 (ported to `skbase` in https://github.com/sktime/skbase/pull/138)

A STEP is available here: https://github.com/sktime/enhancement-proposals/pull/29

Includes:
* test for `get_config` and `set_config` interface

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -76,7 +76,7 @@ class BaseObject(_FlagManager, _BaseEstimator):
 
     def __init__(self):
         """Construct BaseObject."""
-        self._tags_dynamic = {}
+        self._init_flags(flag_attr_name="_tags")
         super(BaseObject, self).__init__()
 
     def __eq__(self, other):
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -77,6 +77,7 @@ class BaseObject(_FlagManager, _BaseEstimator):
     def __init__(self):
         """Construct BaseObject."""
         self._init_flags(flag_attr_name="_tags")
+        self._init_flags(flag_attr_name="_config")
         super(BaseObject, self).__init__()
 
     def __eq__(self, other):
@@ -432,6 +433,38 @@ class BaseObject(_FlagManager, _BaseEstimator):
         self._clone_flags(
             estimator=estimator, flag_names=tag_names, flag_attr_name="_tags"
         )
+
+        return self
+
+    def get_config(self):
+        """Get config flags for self.
+
+        Returns
+        -------
+        config_dict : dict
+            Dictionary of config name : config value pairs. Collected from _config
+            class attribute via nested inheritance and then any overrides
+            and new tags from _onfig_dynamic object attribute.
+        """
+        return self._get_flags(flag_attr_name="_config")
+
+    def set_config(self, **config_dict):
+        """Set config flags to given values.
+
+        Parameters
+        ----------
+        config_dict : dict
+            Dictionary of config name : config value pairs.
+
+        Returns
+        -------
+        self : reference to self.
+
+        Notes
+        -----
+        Changes object state, copies configs in config_dict to self._config_dynamic.
+        """
+        self._set_flags(flag_attr_name="_config", **config_dict)
 
         return self
 
--- a/skbase/tests/conftest.py
+++ b/skbase/tests/conftest.py
@@ -83,7 +83,12 @@ SKBASE_PUBLIC_CLASSES_BY_MODULE = {
     ),
 }
 SKBASE_CLASSES_BY_MODULE = SKBASE_PUBLIC_CLASSES_BY_MODULE.copy()
-SKBASE_CLASSES_BY_MODULE.update({"skbase.base._meta": ("BaseMetaEstimator",)})
+SKBASE_CLASSES_BY_MODULE.update(
+    {
+        "skbase.base._meta": ("BaseMetaEstimator",),
+        "skbase.base._tagmanager": ("_FlagManager",),
+    }
+)
 SKBASE_PUBLIC_FUNCTIONS_BY_MODULE = {
     "skbase.lookup": ("all_objects", "get_package_metadata"),
     "skbase.lookup._lookup": ("all_objects", "get_package_metadata"),
--- a/skbase/tests/test_base.py
+++ b/skbase/tests/test_base.py
@@ -1028,7 +1028,7 @@ def test_set_get_config():
 
     config_start = obj.get_config()
     assert isinstance(config_start, dict)
-    assert set(config_start.keys()) == set(["foo_config", "bar"])
+    assert set(config_start.keys()) == {"foo_config", "bar"}
     assert config_start["foo_config"] == 42
     assert config_start["bar"] == "a"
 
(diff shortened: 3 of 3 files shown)


skbase/base/_tagmanager.py
```diff
@@ -0,0 +1,215 @@
...
+class _FlagManager:
+    """Mixin class for flag and configuration settings management."""
+
+    @classmethod
+    def _get_class_flags(cls, flag_attr_name="_flags"):
+        """Get class flags from estimator class and all its parent classes.
+
+        Parameters
+        ----------
+        flag_attr_name : str, optional, default = "_flags"
```
RNKuhns, line 22: Minor tweak, but [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#sections) notes that if the default value is used then you don't put "optional" part of parameter docstring (e.g., you'd just put flag_attr_name : str, default="_flags").

They also show no spaces around equal sign.

Can you take a quick look at the docstrings in the PR and update these.

skbase/base/_tagmanager.py
```diff
@@ -0,0 +1,215 @@
...
+            name of the flag attribute that is read
+
+        Returns
+        -------
+        collected_flags : dict
```
RNKuhns, line 27: Per [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#returns) if there is only a single return, just the return type can be listed on first line of "returns" description.

RNKuhns: Can you take a look at the returns section of all the docstrings in this PR for conformity with this.

skbase/base/_tagmanager.py
```diff
@@ -0,0 +1,215 @@
...
+        flag_value_default=None,
+        raise_error=True,
+        flag_attr_name="_flags",
+    ):
+        """Get flag value from estimator class and dynamic flag overrides.
+
+        Parameters
+        ----------
+        flag_name : str
+            Name of flag to be retrieved
```
RNKuhns, line 122: The description lines for parameters should end with periods (e.g., Name of flag to be retrieved).

Can you take a look through docstrings in the PR and make sure the periods are there?

RNKuhns: See examples at [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#parameters)

skbase/base/_tagmanager.py
```diff
@@ -0,0 +1,215 @@
...
+
+        Returns
+        -------
+        flag_value :
+            Value of the `flag_name` flag in self. If not found, returns an error if
+            raise_error is True, otherwise it returns `flag_value_default`.
+
+        Raises
+        ------
+        ValueError if raise_error is True i.e. if flag_name is not in self.get_flags(
```
RNKuhns, line 138: Move the description of when the ValueError is raised to the line below the error type (and indent like other descriptions).

skbase/base/_tagmanager.py
```diff
@@ -0,0 +1,215 @@
...
+        Parameters
+        ----------
+        flag_dict : dict
+            Dictionary of flag name : flag value pairs.
+        flag_attr_name : str, optional, default = "_flags"
+            name of the flag attribute that is read
+
+        Returns
+        -------
+        Self : Reference to self.
```
RNKuhns, line 162: Move return type description to line below return type (and remove colon).

skbase/base/_tagmanager.py
```diff
@@ -0,0 +1,215 @@
...
+        estimator : estimator inheriting from :class:BaseEstimator
+        flag_names : str or list of str, default = None
+            Names of flags to clone. If None then all flags in estimator are used
+            as `flag_names`.
+        flag_attr_name : str, optional, default = "_flags"
+            name of the flag attribute that is read
+
+        Returns
+        -------
+        Self :
```
RNKuhns, line 192: Same as above.

RNKuhns (changes requested):
@fkiraly I'm fine with the logic and implementations. But can you go through the docstrings to make sure they follow the numpydoc API.

Otherwise they generate lots of formatting errors/warnings when rendered by sphinx (makes it tough to understand success or failure of actual doc buildingg/formatting)

skbase/base/_base.py
```diff
@@ -449,20 +430,41 @@ def clone_tags(self, estimator, tag_names=None):
...
-            tag_names = tags_est.keys()
-        else:
-            # if tag_set is passed, intersect keys with tags in estimator
-            if not isinstance(tag_names, list):
-                tag_names = [tag_names]
-            tag_names = [key for key in tag_names if key in tags_est.keys()]
+        return self
 
-        update_dict = {key: tags_est[key] for key in tag_names}
+    def get_config(self):
```
RNKuhns, line 439: Not a blocking change, but just noting that later we'll want to generalize `get_config` to get global config and local config (currently only handles local config).

fkiraly: hm, I thought the future logic was going to be, local config overrides global config, and the return is `global.update(local)`

skbase/base/_tagmanager.py
fkiraly: I see, you would use "optional" only if you have sth like `None`. Fixed in all parallel cases

skbase/base/_tagmanager.py
fkiraly: ok, fixed in all instances.

skbase/base/_tagmanager.py
fkiraly: ok

skbase/base/_tagmanager.py
fkiraly: ok, done

skbase/base/_tagmanager.py
fkiraly: same

fkiraly:
I think I've addressed all the above, including parallel locations.

Btw, if we lint `numpydoc`, should we not have it as part of the CI?
I was only loosely following `numpydoc`, deviating where I thought it made things more obscure (e.g., return variable name or "optional")

skbase/base/_base.py
RNKuhns: @fkiraly that's actually what I meant. We'll need an extension point so that we users can specify global config locations that implement something that aligns with the protocol (e.g, can point to their own packages global get_config). The idea would  be to get skbase global configs, user global configs and then local configs. 

Analagous to walking the MRO  and collecting tags, but walking specified global configs.

RNKuhns:
> I think I've addressed all the above, including parallel locations.
> 
> Btw, if we lint `numpydoc`, should we not have it as part of the CI? I was only loosely following `numpydoc`, deviating where I thought it made things more obscure (e.g., return variable name or "optional")

@fkiraly I've opened the issue. And it looks like our thoughts on this are timely. There is a new PR in `numpydoc` repo to allow its validate functionality to be used in a pre-commit hook!

RNKuhns (approved):
LGTM.

PR #142, opened 2023-03-05
[BUG] fix unreported return type bug of `BaseFixtureGenerator.is_excluded`

fkiraly:
This PR fixes an unreported return type bug of `BaseFixtureGenerator.is_excluded` which should return a boolean.

If `self.excluded_tests` is `None`, it returns a list rather than a boolean.

Also adds related detail comments about expected types of class attributes, as comments next to the class attribute definition.


RNKuhns (approved):
Changes make sense. `is_excluded` should now always return boolean.

PR #148, opened 2023-03-11
[MNT] Update slack to point to skbase workspace

RNKuhns:
#### Reference Issues/PRs

This updates the slack link in the documentation to point to the ``skbase`` slack workspace.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -179,7 +179,7 @@ html_theme_options = {
         },
         {
             "name": "Slack",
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Great, makes sense.

One question to consider, whether it's better to have it pointing instead, or also, to `sktime` discord until there are more regular contributors for `skbase`?

RNKuhns:
@fkiraly I think also pointing to sktime discord makes sense.

PR #149, opened 2023-03-13
Add global config interface

RNKuhns:

This adds a global config interface. It allows the retrieval of the global config via ``get_config``, updating the global config via ``set_config``, finding out what ``skbase``'s default config is via ``get_default_config``, resetting the config to the default via ``reset_config``, and a config context manager via ``config_context``.

The PR also updates the local config interface's config retrieval (e.g., `BaseObject().get_config()`) so that it will retrieve the global config and then update it based on local config.

It also adds the ``__skbase_get_config__`` extension point. If it is defined on a descendent class and returns a dict, then the returned dict is used to update the global config. This is useful for downstream packages that use ``skbase`` to implement their own extension to ``skbase`` that reads their own package's global config.

Order of precedence is:

1. Retrieve a copy of the global config
2. If ``__skbase_get_config__`` is defined and returns a dict, then it is used to update the copy of the global config
3. Use local config interface to update the copy of the global config

There is not any new dependencies added.

Does design make sense? Do tests cover any cases we can think of?

##### For all contributions

##### For code contributions

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -63,6 +63,7 @@ from sklearn.base import BaseEstimator as _BaseEstimator
 
 from skbase._exceptions import NotFittedError
 from skbase.base._tagmanager import _FlagManager
+from skbase.config import get_config
 
 __author__: List[str] = ["mloning", "RNKuhns", "fkiraly"]
 __all__: List[str] = ["BaseEstimator", "BaseObject"]
@@ -446,7 +447,28 @@ class BaseObject(_FlagManager, _BaseEstimator):
             class attribute via nested inheritance and then any overrides
             and new tags from _onfig_dynamic object attribute.
         """
-        return self._get_flags(flag_attr_name="_config")
+        config = get_config().copy()
+
+        # Get any extension configuration interface defined in the class
+        # for example if downstream package wants to extend skbase to retrieve
+        # their own config
+        if hasattr(self, "__skbase_get_config__") and callable(
+            self.__skbase_get_config__
+        ):
+            skbase_get_config_extension_dict = self.__skbase_get_config__()
+        else:
+            skbase_get_config_extension_dict = {}
+        if isinstance(skbase_get_config_extension_dict, dict):
+            config.update(skbase_get_config_extension_dict)
+        else:
+            msg = "Use of `__skbase_get_config__` to extend the interface for local "
+            msg += "overrides of the global configuration must return a dictionary.\n"
+            msg += f"But a {type(skbase_get_config_extension_dict)} was found."
+            warnings.warn(msg, UserWarning, stacklevel=2)
+        local_config = self._get_flags(flag_attr_name="_config")
+        config.update(local_config)
+
+        return config
 
     def set_config(self, **config_dict):
         """Set config flags to given values.
--- a/skbase/config/__init__.py
+++ b/skbase/config/__init__.py
@@ -1,0 +1,404 @@
+# -*- coding: utf-8 -*-
+""":mod:`skbase.config` provides tools for global configuration of ``skbase``."""
+# -*- coding: utf-8 -*-
+# copyright: skbase developers, BSD-3-Clause License (see LICENSE file)
+# Includes functionality like get_config, set_config, and config_context
+# that is similar to scikit-learn. These elements are copyrighted by the
+# scikit-learn developers, BSD-3-Clause License. For conditions see
+# https://github.com/scikit-learn/scikit-learn/blob/main/COPYING
+import collections
+import sys
+import threading
+import warnings
+from contextlib import contextmanager
+from dataclasses import dataclass
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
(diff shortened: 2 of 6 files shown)


RNKuhns:
@fkiraly I have to investigate the two test failures. They are only tangentially related to the global config functionality. When testing the metadata for the module it is comparing ``skbase.config.GlobalConfigParamSetting`==skbase.config.GlobalConfigParamSetting` `and returning False, which is leading to the test failures.

Not exactly sure why that is happening as I'm having trouble replicating it outside the usage of the metadata retrieval function. 

While I troubleshoot that, can you take a first look at the interface and provide feedback.

RNKuhns:
@fkiraly I plan on adding a few more simple tests of uncovered edge cases to increase coverage. But this is ready for feedback in the interim.

fkiraly (changes requested):
Hm - looks good but I am not sure I fully understand the design.

If this were a downstream package, I would be agreed.

But what ist the extension case here? That is, someone using `skbase` to make a package.

I think this is going to be problematic, as it now ties every descendant of `BaseObject` to the global config in `skbase`.

Instead, I would invert the pointer, say, add a method `get_global_config` or similar that by default does nothing in the downstream package, but can be linked to a global config in the downstream package (which in turn is templated by `skbase`).

RNKuhns:
@fkiraly good questions.


> If this were a downstream package, I would be agreed.
> But what ist the extension case here? That is, someone using `skbase` to make a package.
> I think this is going to be problematic, as it now ties every descendant of `BaseObject` to the global config in `skbase`.

There are two things at play:

1. Reliance of downstream packages to the config or options of an upstream package.
2. Should we let users override global ``skbase`` functionality at the instance level via `BaseObject.set_config`

I've included more details on both below.

### Reliance of downstream packages to the config or options of an upstream package.
There are multiple cases where downstream Python packages need to call an upstream package's config/options to change behavior in a way that affects downstream (I'm showing `scikit-learn` example, but other packages like `pandas`, `dask`, `ray`, etc. all have config/options/settings that change how things run in downstream packages that use them.

For example, packages that use `scikit-learn`'s `BaseEstimator` already are tied to having to set the global config for `scikit-learn` to change the behavior of `BaseEstimator`.  The code example below shows a stylized pretty printing example.

For example:

```python
from sklearn import set_config
from sklearn.base import BaseEstimator

class SomeDownstreamPackageClass(BaseEstimator):
    """This class implements some awesome estimator."""
    
    def __init__(self, some_param=7):
        self.some_param = some_param

downstream_estimator = SomeDownstreamPackageClass()
print(downstream_estimator)
# prints SomeDownstreamPackageClass()

set_config(print_changed_only=False)  # True is default
print(downstream_estimator)
# prints SomeDownstreamPackageClass(some_param=7)
```
The first part of the global config interface (part importable from `skbase.config`) essentially adds the ability to global configure how `skbase` code artifacts should perform (for now it is just pretty printing but could be more in future).

But this has the downside that it means the local `BaseObject.set_config` does not affect the instances pretty printing behavior. 
As it stands now, if an `sktime` or `skbase` class that has a local `set_config` method called it, that would not be able to override the `scikit-learn` global configuration. The second part of the PR that ties into `BaseObject.get_config` ensures that this is possible. There isn't really a cost here, since we get a copy of the `skbase` config and then potentially override it. The packages are already depending on the `skbase` config because it controls how the pretty printing works, so we are really just giving them extra flexibility to override it at the instance level.

### Should we let users override global ``skbase`` functionality at the instance level via `BaseObject.set_config`

I think a `skbase.BaseObject` should be able to return all the configurations that impact how it behaves. All `BaseObject`s are affected by the `skbase` configuration for how they should be pretty printed. So any descendent of `BaseObject` should be able to report the dependency via `BaseObject.get_config` and override the behavior locally on the object via `BaseObject.set_config`. Users who want to override the behavior globally for all their descendent objects can use the `skbase.config.set_config` interface similar to how a user would override pretty printing behavior of `sktime` objects using `scikit-learn`'s `set_config`.

The question about how downstream users can extend the `BaseObject` to allow overrides of their own global interface is discussed in the next section.

> Instead, I would invert the pointer, say, add a method `get_global_config` or similar that by default does nothing in the downstream package, but can be linked to a global config in the downstream package (which in turn is templated by `skbase`).

This is similar to what the behavior of the ``__skbase__get_config__`` dunder method does in the PR. If it is not present then nothing occurs. But if it is present (and returns a dictionary) then it is assumed to be the entry point for downstream users to provide their global config to the `BaseObject`. I'm open to it being a `BaseObject` method that we default to returning an empty dict and tell downstream users to override for this behavior. But I also think we should consider using a ``__skbase_some_functionality__`` dunder method interface for this and other extension points (another point is letting downstream packages provide custom cloning logic to be used by `clone` method, for cases when special consideration need to be given to certain parameter values beyond that done for the random number generators. In this case a non-public method would also work).

But from a configuration perspective, all objects in a downstream class are potentially affected by: 

1. `skbase` global config (pretty printing currently. Potentially other behavior in the future)
2. global config of downstream package
3. Local user config on instance

The `BaseObject.get_config` as set up in the PR retrieves the configurations in that order. `BaseObject.set_config` lets users change the local intance config, but not affect the `skbase` global config  or the downstream package config.

PR #150, opened 2023-03-13
[ENH] Implements sklearn style pretty printing directly in ``skbase``

### LINKED ISSUE #17: [ENH] Separate Pretty Printing from sklearn dependency
A clear and concise description of what the problem is. 

Eventually the `baseobject` package will not depend on scikit-learn's `BaseEstimator`. This will be accomplished by gradually removing the `BaseObject` dependencies on `BaseEstimator` for different API interfaces and features.

This issue covers the features that need to be added to `BaseObject` to remove the dependency on `BaseEstimator` when pretty printing instances of classes that inherit from `BaseObject`.

A clear and concise description of what you want to happen, ideally taking into consideration the existing toolbox design, classes and methods. 

`BaseObject` should contain all the functionality needed to implement the pretty printing. However, we need to investigate and eventually decide the degree to which the functionality in `BaseObject` should identically recreate the pretty printing functionality present in `BaseEstimator`.

Important considerations are:

- Ease of maintence
  - There may be ways we can get a similar effect with less code to maintain, but we need to understand how the pros and cons associated with any different implementations of this functionality
- How `BaseObject` interacts with `sklearn` classes/pipelines when pretty printing representations of classes
- Need for adequate tests to ensure pretty printing is working correctly and meets the level of compatability with `sklearn` that we choose

RNKuhns:
#### Reference Issues/PRs

Fixes #17

This implements ``sklearn`` pretty printing directly in ``skbase`` with some minor tweaks.

No new dependencies. In theory this will let us remove the dependency on ``sklearn``, but we'll save that for a separate PR.

This depends on the global config PR -- specifically, the ability for local config interface to retrieve (and update) global config. Two global (or local) configuration parameters (print_changed_only and display) control pretty printing output.

##### For all contributions

##### For code contributions


RNKuhns:
@fkiraly I want to do a bit more testing and I need to clean up the 2 code scanning items that were found, but this should be ready for you to take a look at.

fkiraly (changes requested):
Quick question: why is this dependent on the global config?

This could also be implemented with, and tested with, a purely local config mechanism?

RNKuhns:
@fkiraly this could in theory be implemented as a local configuration. But in practice, how the objects are printed or represented in html is really something you want to set and have apply to all the objects.

fkiraly:
no, I meant architecturally.

Why does this PR need to depend on the global config?
Should the global config not be such that it can optionally set a global default for any local config?

In my (perhaps too naive) understanding of the global config, it should not be coupled with the pretty printing config option, so it should not be necessary to have this PR depend on the global config.

Conversely, if there is a strict dependency here, I think there's something wrong architecturally (perhaps subtly)

RNKuhns:
@fkiraly i think you are suggesting the local config is the  main/prevailing config and the global config can optionally override it.

That would imply returning local config and then updating with global config.

However, I think the global config should be the higher precedent config and the local config should be to optionally override it.

The global configuration should only contain config params and values that provide broad context to a BaseObject.

Local config can override that context for an object. 

So for something like pretty printing options. The most typical use case is that they are not changed from global config settings. So in a limited number of cases a user calls the object's set_config to override global setting.

fkiraly:
@RNKuhns, I am not suggesting anything about precendence of configs here (and I agree with you that local should override global).

All I'm saying that the existene of a global config mechanism should not be architecturally required for the pretty printing. If it is, something is architecturally wrong.

Why: config precedence and "getting the values for the applicable config" should be architecturally separate to translate the applicable config values into behaviour.

fkiraly:
supserseded by https://github.com/sktime/skbase/pull/156. Config interface is here: https://github.com/sktime/skbase/pull/149

Issue #145, opened 2023-03-08
Pydata Work Prioritization


RNKuhns:
This is a meta issue to prioritize our work for Pydata Seattle 2023.

**High-level tasks for the presentation:**
- [x] #141

High-level tasks for new features/maintenance ahead of Pydata Seattle 2023. These are divided into "must", "should" and "nice-to-have":

**Must:**
- [x] `sktime` and `skbase` initial integration is completed. Targeting mid-April release of `sktime` v0.18.0
- [x] Create "example" repository that includes a stylized package that uses `skbase` (will also be useful for additional testing of our lookup, testing and validation functionality too)
- [x] Local (class) configuration interface
- [x]  #106 

**Should**
- [x] #150 
- [ ] Add persistance (re-work of `sktime` persistance approach to make sure it works more generally)
- [ ] Upgrade documentation, with emphasis on landing page
- [ ] Redo governance section of documentation
- [ ] Create functionality roadmap (borrow from HackMD and design document and add other ideas)
- [ ]  Improve documentation of the `skbase.testing` module

**Nice-to-have**
- [ ] #149 
- [ ] Example showing how to use `skbase` to create `scikit-learn` compliant class (need to investigate if this is possible. If their checks are "duck-typed" then it should be)

fkiraly:
seems like you dropped from the call?

My thoughts on priorities:

1. "cook show" style tutorial creation, as described here: https://github.com/sktime/skbase/issues/141
consists of two parts: (1a) jupyter notebooks for demo; (1b) multiple stages of "package being built" for the cook show

2. `sktime`/`skbase` integration. It is realistic to publish an `skbase` based `sktime` version 0.18.0 by mid Apr - test framework refactor fails for mysterious reasons, but it is not actually needed for a rebasing of the core module, as the existing `sktime` tests are robust enough.

3. optional bonus feature: persistence. This has been implemented in `sktime` by myself and @achieveordie, and is likely a nice feature we could deliver until then, in `skbase` (lateral transfer)

4. optional proof-of-concept: `sklearn` with `skbase`. Small version: extension template based on `skbase` that passes `sklearn` interface contract tests. Big version: PR to `sklearn` fork that replaces the base class and refactors the tag system to use `skbase`'s, etc.

fkiraly:
from call with Ryan - which features would be great to create that we can showcase?
These are "should", "must" is the actual material based on status quo.

* config (local) - PR exist
* global config - may need some thought
[...]

PR #154, opened 2023-04-02
[ENH] Introduce Serialization Support

achieveordie:
Part of #145 - Introduces serialization functionality.

Currently, this is a direct port from the functionality we have in `sktime`. I'll iteratively improve the code as per requirements and inputs. 

- [x] Port serialization methods (`load` as a standalone function, `save` as `BaseObject` method) from `sktime`
- [ ] Add tests to ensure the functionalities added work properly and are consistent with the expected behaviour.
- [ ] Improve/add code to ensure this functionality can be generally applied and extended.

None

- since `bandit` complains about the usage of pickle. I've excluded two tests ([import](https://bandit.readthedocs.io/en/0.0.0/blacklists/blacklist_imports.html#b403-import-picle) and [load](https://bandit.readthedocs.io/en/0.0.0/blacklists/blacklist_calls.html#b301-pickle)) and will add a warning in `load`'s docstring regarding it. We should try to find safer alternatives to replace it in the future.
- What improvements should we bring to the current code? Some nice-to-have functionalities etc

##### For all contributions

##### For code contributions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,6 +141,7 @@ ignore_path = ["docs/_build", "docs/source/api_reference/auto_generated"]
 
 [tool.bandit]
 exclude_dirs = ["*/tests/*", "*/testing/*"]
+skips = ["B301", "B403"]
 
 [tool.setuptools]
 zip-safe = true
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -682,6 +682,88 @@ class BaseObject(_FlagManager, _BaseEstimator):
 
         return comp_dict
 
+    def save(self, path=None):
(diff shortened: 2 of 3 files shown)


fkiraly (changes requested):
Looks great!

Before I go into detail, I wanted to say that the tests are failing because `skbase` uses itself as a test case for the crawling utilities, and requires you to enter new files, modules, classes, etc, in `tests.conftest` (the lists).

Personally, I think this is very tedious for a contributor and wasn't a great design decision, so mid-term we need to get away from that. For now, I'd recommend you just add to the lists.

fkiraly:
let me know when this is ready

PR #178, opened 2023-05-04
& yarnabrina [MNT] move `deep_equals` and dependency checkers from `testing` to `utilities` to remove accidental coupling to `pytest`

### LINKED ISSUE #177: [BUG] `clone` breaks when `skbase` is installed without `pytest`
Calling an estimator's `clone` breaks  when `pytest` is not installed, due to accidental coupling to `pytest` as discussed here: https://github.com/sktime/sktime/issues/4544#issuecomment-1534505500

fkiraly:
This PR moves `deep_equals` and dependency checkers from `testing` to `utilities` to remove accidental coupling to `pytest`.

Also see here for the problem caused by this: https://github.com/sktime/skbase/issues/177 - this fixes #177

Crediting @yarnabrina for the fix as suggested in https://github.com/sktime/sktime/issues/4544#issuecomment-1534505500


yarnabrina (commented):
I know almost nothing about how `skbase` works, but is it not necessary to add `deep_equals` to `__init__.py`? Or, it'd be this way during deprecation period, and then changed?

yarnabrina:
> Unfortunately the module name is the same as the function name, so it would result in a clash to import the function to the same module that contains the module of the same name, no?

Just to be clear, I meant doing something like this in `skbase/utils/__init__.py`:

```python3
## existing imports

from sktime.utils.deep_equals import deep_equals

## or if you accept relative imports
## from .deep_equals import deep_equals

## existing codes
```

Tried locally, and it works.

```pycon
>>> import skbase.utils
>>> skbase.utils.deep_equals
<function deep_equals at 0x7f0266974ee0>
>>>
```

Can you please check https://github.com/sktime/skbase/pull/181?

PR #190, opened 2023-05-19
[MNT] address deprecation of `load_module`

fkiraly:
`load_module` is deprecated and will be replaced by `exec_module` in python 3.12, this PR addresses that.

The fix is highly nontrivial - the crucial line was
`imported_mod = module.load_module()` for a `module: SourceFileLoader`,

which had to be replaced by the equivalent four (!) lines

```python
spec = importlib.util.spec_from_file_location(module.name, module.path)
imported_mod = importlib.util.module_from_spec(spec)

# these two are also crucial due to the side effects
loader = spec.loader
loader.exec_module(imported_mod)
```

The deprecation message raised by `load_module` is highly unhelpful, and I had to reverse engineer this from a very helpful PR in `nox`, https://github.com/wntrblm/nox/pull/498

### DIFF
--- a/skbase/lookup/_lookup.py
+++ b/skbase/lookup/_lookup.py
@@ -297,7 +297,12 @@ def _import_module(
         if isinstance(module, str):
             imported_mod = importlib.import_module(module)
         elif isinstance(module, importlib.machinery.SourceFileLoader):
-            imported_mod = module.load_module()
+            spec = importlib.machinery.ModuleSpec(
+                name=module.name,
+                loader=module,
+                origin=module.path,
+            )
+            imported_mod = importlib.util.module_from_spec(spec)
         exc = None
     except Exception as e:
         # we store the exception so we can restore the stdout fisrt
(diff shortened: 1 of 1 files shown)


fkiraly:
FYI @yarnabrina - perhaps you see a quick fix to this? I'm not sure what to replace the deprecated logic by, is the problem.

yarnabrina:
Sorry, I have not used `exec_module` ever. But the documented code seems be in a different format (for example, with arguments).

https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly

If you can explain me what you are trying to do, for example why the `if` block is not enough and it has to go to `eif` block, I can try to think on this. Given this is a internal function, can you not control what is passed to this function?

fkiraly:
Re the "what", the issue is simply that `load_module` is deprecated, and I am trying to preserve the logic of the affected code block by any means, without using `load_module`. There are no further constraints.

fkiraly:
for posterity, this was highly non-trivial.
I reverse-engineered the solution from https://github.com/wntrblm/nox/pull/498.

Any line `imported_mod = module.load_module()` for a `module: SourceFileLoader` can equivalently be replaced by apparently equivalent the four lines

```python
spec = importlib.util.spec_from_file_location(module.name, module.path)
imported_mod = importlib.util.module_from_spec(spec)

# these two are also crucial due to the side effects
loader = spec.loader
loader.exec_module(imported_mod)
```

The deprecation message raised by `load_module` is highly unhelpful here.

Issue #208, opened 2023-07-13
[MNT] windows CI failing in `conda` setup step


fkiraly:
The windows CI seems to be failing with a `conda` related error:

![image](https://github.com/sktime/skbase/assets/7985502/8cfc7f31-9132-449e-99a0-775185568251)

Oddly, the `sktime` CI which should be identical is not failing.

PR #209, opened 2023-07-13
[MNT] separate windows CI element from unix based CI

fkiraly:
Fixes https://github.com/sktime/skbase/issues/208, by replicating the `sktime` setup, which separates windows CI and unix based CI.


fkiraly:
Hm, this seems to fix it, and the tests all run.

PR #275, opened 2024-01-30
[BUG] Fix cloning of config for nested objects

tpvasconcelos:
#### Reference Issues/PRs

Fixes https://github.com/sktime/skbase/issues/276

Please refer to: https://github.com/sktime/skbase/issues/276

No

The `_clone()` method has a `# copied from sklearn` comment above it. Any concerns with modifying this method? Any alternative solutions come to mind?

##### For all contributions

##### For code contributions

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -225,7 +225,14 @@ class BaseObject(_FlagManager):
         estimator_type = type(estimator)
         # XXX: not handling dictionaries
         if estimator_type in (list, tuple, set, frozenset):
-            return estimator_type([self._clone(e, safe=safe) for e in estimator])
+            return estimator_type(
+                [
+                    e.clone()
+                    if isinstance(e, BaseObject)
+                    else self._clone(e, safe=safe)
+                    for e in estimator
+                ]
+            )
         elif not hasattr(estimator, "get_params") or isinstance(estimator, type):
             if not safe:
                 return deepcopy(estimator)
@@ -247,7 +254,11 @@ class BaseObject(_FlagManager):
         klass = estimator.__class__
         new_object_params = estimator.get_params(deep=False)
         for name, param in new_object_params.items():
-            new_object_params[name] = self._clone(param, safe=False)
+            new_object_params[name] = (
+                param.clone()
+                if isinstance(param, BaseObject)
+                else self._clone(param, safe=False)
+            )
         new_object = klass(**new_object_params)
         params_set = new_object.get_params(deep=False)
 
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -160,9 +160,6 @@ class BaseObject(_FlagManager):
         self_params = self.get_params(deep=False)
         self_clone = self._clone(self)
 
-        if self.get_config()["clone_config"]:
-            self_clone.set_config(**self.get_config())
-
         # if checking the clone is turned off, return now
         if not self.get_config()["check_clone"]:
             return self_clone
@@ -225,14 +222,7 @@ class BaseObject(_FlagManager):
         estimator_type = type(estimator)
         # XXX: not handling dictionaries
         if estimator_type in (list, tuple, set, frozenset):
-            return estimator_type(
-                [
-                    e.clone()
-                    if isinstance(e, BaseObject)
-                    else self._clone(e, safe=safe)
-                    for e in estimator
-                ]
-            )
+            return estimator_type([self._clone(e, safe=safe) for e in estimator])
         elif not hasattr(estimator, "get_params") or isinstance(estimator, type):
             if not safe:
                 return deepcopy(estimator)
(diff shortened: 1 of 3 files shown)


fkiraly (changes requested):
Astutely observed.

However, question, given that there are two points of recursion

* the first, this deals with nesting in lists etc.
* the second, `new_object_params[name] = self._clone(param, safe=False)`, this deals with direct nesting.

Naively, I would think we do *not* need to change the first occurrence, current line 228, because we go back to `_clone` and from there to line current 250.

Further, non-blocking question, I was wondering: should we make the contract less strict about inheritance, and more about interfaces, would it make sense to do `hasattr(param, "clone") and hasattr(param, "_config")`? Or perhaps it is safer to make the *deviation* from `sklearn` more strict, which is the current state. I tend towards the latter.

skbase/tests/test_base.py
```diff
@@ -966,6 +966,45 @@ class TestClass(BaseObject):
...
+
+    class TestClass(BaseObject):
+        _tags = {"some_tag": True, "another_tag": 37}
+        _config = {"check_clone": 0}
+
+        def __init__(self, nested=None):
+            self.nested = nested
+
+    test_obj = TestClass(
+        nested=[TestClass().set_config(**{"check_clone": 42, "foo": "bar"})]
```
fkiraly, line 981: we should also check `nested` being another estimator (not wrapped in a list) - I think that would fail with th current fix, see above.

skbase/base/_base.py
```diff
@@ -260,6 +257,11 @@ def _clone(self, estimator, *, safe=True):
                     "Cannot clone object %s, as the constructor "
                     "either does not set or modifies parameter %s" % (estimator, name)
                 )
+
+        # This is an extension to the original sklearn implementation
+        if isinstance(estimator, BaseObject) and estimator.get_config()["clone_config"]:
```
fkiraly, line 262: I assume this is now WiP, just wanted to mention that anything you move after the lines
```python
        if not self.get_config()["check_clone"]:
            return self_clone
```
will not happen if `check_clone` is off, and it is off by default.

tpvasconcelos: I think that you misread the change. This change is inside `_clone()` which is called **before** that _if_-block. Does that make sense?

skbase/tests/test_base.py
tpvasconcelos: Good point! I updated the test to also include an estimator parameter. I think that test now covers the behaviour we need to cover here however I would like to refactor it as the implementation is quite _ugly_ currently. Maybe this could be a different PR where we clean-up/refactor the `test_config_after_clone_tags` and `test_nested_config_after_clone_tags` tests

tpvasconcelos:
> Naively, I would think we do _not_ need to change the first occurrence, current line 228, because we go back to `_clone` and from there to line current 250.

I changed the implementation to just do the config cloning inside the `_clone` method instead of having it in the `clone` method. I think this is a cleaner implementation.

> Further, non-blocking question, I was wondering: should we make the contract less strict about inheritance, and more about interfaces, would it make sense to do hasattr(param, "clone") and hasattr(param, "_config")? Or perhaps it is safer to make the deviation from sklearn more strict, which is the current state. I tend towards the latter.

I don't have a strong opinion here but I can't see why someone wouldn't want to clone the config of an object without cloning the config of "nested" objects. Can you? Maybe this can be tackled when/if someone flags it as an issue? i.e., I think that the default behaviour should be that the config gets cloned for all inner/nested objects 

WDYT?

tpvasconcelos:
> Further, non-blocking question, I was wondering: should we make the contract less strict about inheritance, and more about interfaces, would it make sense to do `hasattr(param, "clone") and hasattr(param, "_config")`? Or perhaps it is safer to make the _deviation_ from `sklearn` more strict, which is the current state. I tend towards the latter.

With the current implementation (see latest changes) this is not relevant anymore. Could you take a look? 🙏

skbase/base/_base.py
fkiraly: oh, yes, I must have somehow hallucinated the change somewher else. You are right.

fkiraly (approved):
Very elegant!

Putting the `set_config` at the recursion nodes rather than having it just in the top node solves the problem conclusively and elegantly. Nice.

Cohesion is maintained since the total number of calls does not change, and `_clone` gets to handle `set_params` and `set_config`, so even that is improved.

skbase/tests/test_base.py
fkiraly: tests don't need to win beauty contests, they need to be correct, easy to understand, and cover relevant cases, all of which I belive is true even right now.

However, I see where you are coming from, so by all means!

fkiraly:
You should also add yourself here: https://github.com/sktime/skbase/blob/main/.all-contributorsrc

`bug`, `code`, `ideas`.

fkiraly:
> The `_clone()` method has a `# copied from sklearn` comment above it. Any concerns with modifying this method?

Since you are asking: in-principle no concerns. The history behind `_clone` is that we originally wanted a clean re-implementation which turned out to be discrepant from `sklearn` in some important examples due to the magic after `if estimator_type in (list, tuple, set, frozenset):` and `safe` arg which defines a very specific contract.

Ultimately, one requirement for `BaseObject` is that `BaseEstimator` could replace `sklearn` `BaseEstimator` with at most minimal adjustments, so the only quick way to see how to get there was to copy-paste `sklearn` `clone`.

Funnily, `sklearn` seem to be going in a similar direction by allowing estimators to dispatch to their internal `clone` (which seems a bit unsafe to me compared to base class inheritance). Either way, some tech discussions on uniformization of APIs are going to be scheduled for later this year between the two projects (`sktime`, `sklearn`) - in case you are interested.

PR #293, opened 2024-02-23
[DOC] Replace use of "estimator" term in base object interfaces with more general references

tpvasconcelos:
#### Reference Issues/PRs

Depends on https://github.com/sktime/skbase/pull/281

See https://github.com/sktime/skbase/pull/281#issuecomment-1956646894 and relevant replies.

##### For all contributions

##### For code contributions

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -169,6 +169,8 @@ def _check_clone(original, clone):
             f"Error on equality check of arguments (x) vs parameters (y): {msg}"
         )
 
+    return clone
+
 
 class BaseObject(_FlagManager):
     """Base class for parametric objects with sktime style tag interface.
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -1324,19 +1324,13 @@ class BaseEstimator(BaseObject):
         if not deep:
             return fitted_params
 
-        def sh(x):
-            """Shorthand to remove all underscores at end of a string."""
-            if x.endswith("_"):
-                return sh(x[:-1])
-            else:
-                return x
-
         # add all nested parameters from components that are skbase BaseEstimator
         c_dict = self._components()
         for c, comp in c_dict.items():
             if isinstance(comp, BaseEstimator) and comp._is_fitted:
                 c_f_params = comp.get_fitted_params(deep=deep)
-                c_f_params = {f"{sh(c)}__{k}": v for k, v in c_f_params.items()}
+                c = c.strip("_")
+                c_f_params = {f"{c}__{k}": v for k, v in c_f_params.items()}
                 fitted_params.update(c_f_params)
 
         # add all nested parameters from components that are sklearn estimators
@@ -1349,7 +1343,8 @@ class BaseEstimator(BaseObject):
             for c, comp in old_new_params.items():
                 if isinstance(comp, self.GET_FITTED_PARAMS_NESTING):
                     c_f_params = self._get_fitted_params_default(comp)
-                    c_f_params = {f"{sh(c)}__{k}": v for k, v in c_f_params.items()}
+                    c = c.strip("_")
+                    c_f_params = {f"{c}__{k}": v for k, v in c_f_params.items()}
                     new_params.update(c_f_params)
             fitted_params.update(new_params)
             old_new_params = new_params.copy()
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -160,9 +160,6 @@ class BaseObject(_FlagManager):
         self_params = self.get_params(deep=False)
         self_clone = self._clone(self)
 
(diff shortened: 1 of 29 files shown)


skbase/base/_base.py
```diff
@@ -1323,19 +1324,13 @@ def get_fitted_params(self, deep=True):
...
-            else:
-                return x
-
         # add all nested parameters from components that are skbase BaseEstimator
         c_dict = self._components()
         for c, comp in c_dict.items():
             if isinstance(comp, BaseEstimator) and comp._is_fitted:
                 c_f_params = comp.get_fitted_params(deep=deep)
-                c_f_params = {f"{sh(c)}__{k}": v for k, v in c_f_params.items()}
+                c = c.strip("_")
```
fkiraly, line 1332: two questions about this:

* this is not part of #281, is that intentional? This is not about consistent use of "estimator" either.
* this is not a refactor but a change in logic, hence potentially breaking dependent code. `sh` removed only a single trailing underscore, whereas `strip` removes both an arbitrary amount of trailing underscores, as well as leading underscores - the latter could be a consequential, breaking change, even in `sktime` I remember examples of things like `_sthsth_` as attribute names

Irrespective of what we think the "should state" should be, we should not make breaking changes without deprecation.

tpvasconcelos: Hey @fkiraly this was just some housecleanning indeed and not related to anything else. You are correct about the removal of leading underscores! That's my bad! I should've used `rstrip()` to only strip trailing underscores. Btw, the `sh()` helper also strips arbitrary trailing underscores (it uses recursion).

```pycon
>>> def sh(x):
...     """Shorthand to remove all underscores at end of a string."""
...    if x.endswith("_"):
...        return sh(x[:-1])
...    else:
...        return x    
>>> a = "_foo__bar____"
>>> sh(a)
'_foo__bar'
>>> a.rstrip("_")
'_foo__bar'
>>> sh(a) == a.rstrip("_")
True
>>> b = "___foo__bar"
>>> sh(b) == b.rstrip("_")
True
```

 Would you be okay with keeping the change but fixing it by using the `rstrip` method? The only reason I did it was because `sh()` is kinda reinventing the wheel since `rstrip` does exactly the same and seems a bit more idiomatic.
 
 I don't mind either way! 👍

tpvasconcelos: Just pushed a commit in case...

fkiraly: > Would you be okay with keeping the change but fixing it by using the `rstrip` method?

Yes - that is equivalent, up to and including the failure if the input is not a `str`. So it would be a refactor, and an improvement.

tpvasconcelos:
@fkiraly I think this is ready to merge/review 👍

skbase/base/_base.py
```diff
@@ -67,6 +67,109 @@
...
+    object doesn't have a `random_state` parameter), an *exact clone* is
+    returned: the clone and the original object will give the exact same
+    results. Otherwise, *statistical clone* is returned: the clone might
+    return different results from the original object. More details can be
+    found in :ref:`randomness`.
+    """
+    obj_type = type(obj)
+    # XXX: not handling dictionaries
+    if obj_type in (list, tuple, set, frozenset):
+        return obj_type(_clone(e, safe=safe) for e in obj)
```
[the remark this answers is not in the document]
tpvasconcelos: Same as: https://github.com/sktime/skbase/pull/281#discussion_r1500621807

fkiraly (changes requested):
Mostly agreed, I have one issue that imo we need to resolve.

In some occurrences, I anticipate that using "object" to refer to a class - namely one inheriting from `BaseObject` - will cause a lot of confusion.

This is because there are python objects - instances of classes - so we are overloading a term that has a much more general meaning.

The most risky instances of these are the case of "object class" or "base object class", because the thing in question is neither a class, nor a base class..

These occcurences - in `set_tag`, `get_tag`, `get_tags`, `set_params`, `get_params` I would leave as is, unless we find a better solution. Perhaps we can work with "from `self`" etc, but that might confuse users who aren't too familiar wit the `self` concept.

Some solution suggestions:

* `get_tags`, `get_tag` - just say "get tags from object and dynamic tag overrides" etc. Or change it completely, to "get tag value, with key-wise inheritance", then explain inheritance in another sentence.
* `get_class_tags`, `get_class_tag` - "from class, with key-wise inheritance"?

Other comments:

* `create_test_intsance`, does not create an instance of `BaseObject`, but of `cls`
* "sub-estimator" sounds strange. Component objects?

fkiraly:
@tpvasconcelos, are you still working on this?

fkiraly (approved):
Made changes to clarify docs; reverted changes to `_clone` to avoid a divergence from `sklearn`.

PR #347, opened 2024-08-02
[MNT] Move release CI to macos-12 image

szepeviktor:
v11 is not available any more
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#standard-github-hosted-runners-for-public-repositories

@fkiraly those jobs do not start
https://github.com/sktime/skbase/actions/runs/10212328626


fkiraly:
I see, thanks!

fkiraly (approved):

PR #372, opened 2024-10-07
[MNT] full support for python 3.13

fkiraly:
Python 3.13 was released today. This PR adds full support for python 3.13:

* removes markers to skip sklearn incompatibility with rc
* changes tests to use 3.13 and not the rc


fkiraly:
3.13 seems not yet available on GHA.

But soon...

PR #380, opened 2024-10-30
[ENH] customizable special attrs in clone

fkiraly:
This PR adds a `clone_attrs` config field which lets users specify a list of strings, attribute names which are retained in `clone` beyond the configs and init parameters.

This is by default `None`, so retains current behaviour.

An approach to https://github.com/sktime/sktime/issues/7333, which would be solved if we set the `sklearn` specific attribute responsible for `set_output` in this list, in `sktime` only.

The main drawback of this would be that `sklearn` compatibility is not automatic, but contingent on setting this hard-to-find config.


fkiraly:
superseded by #381 and #383

Issue #391, opened 2025-01-05
[BUG] Contributors page is broken


fkiraly:
The `Contributors.md` page with account thumbnails and emoji keys is broken.

Currently, the file does not seem to be updated, and nothing is displayed (e.g., no list of contributors).

We should:

* ensure that it it works again - the setup can be copied from `sktime`, since `skbase` is also using the `all-contributors` specification.
* optimally, contributors are displayed at the bottom of the README, similar to `sktime`.

Good first issue, as it can probably be resolved by copying from `sktime`.

Spinachboul:
@fkiraly 
I have raised a PR for the same here: https://github.com/sktime/skbase/pull/394
Kindly review it! Thanks..

PR #394, opened 2025-01-16
 Update Contributors List and Badge

Spinachboul:
Fixes https://github.com/sktime/skbase/issues/391
* Updated the contributors list with new additions.
* Included individual contribution counts for each contributor.
* Added the all-contributors badge for better tracking of contributions


fkiraly (changes requested):
Thanks! Really great!

* can you remove the bots, kindly? Like `dependabot`.
* could you describe how you did this, e.g., executing sth on command line?

What would be very useful is a CI step to automate the update, like in `sktime`. Though that does not need to be done in this PR.

Spinachboul:
> Thanks! Really great!
> 
> * can you remove the bots, kindly? Like `dependabot`.
> * could you describe how you did this, e.g., executing sth on command line?
> 
> What would be very useful is a CI step to automate the update, like in `sktime`. Though that does not need to be done in this PR.

Yes! I would need some guidance on doing that. Though I'd love implementing that here

fkiraly (approved):
Thanks!

I made some minor changes:

* reverted diacritic replacements - probably accidental?
* I replaced the new paragraph in the README with a link to the contributors page. Your proposed change had a couple duplicated paragraphs, and a fixed list which would not update with new contributors.
    * I think this is a good idea, but we should link only when we have figured out how to keep the list updated, and using the emoji key.

PR #396, opened 2025-01-24
Updating the main branch of the forked repository for the recently made commits in the main codebase

Spinachboul:
Refereces https://github.com/sktime/skbase/issues/391

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -131,6 +131,16 @@
         "bug",
         "code"
       ]
+    },
+    {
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
I am slightly confused about this Pr - what does it do, and why is it necessary?

Spinachboul:
@fkiraly 
This PR seems to be updating the `readme` file, but seems that is already included in the main codebase.
I was thinking of opening another PR where we could mention the type of contributions of the contributors instead of the number of contribution

Issue #406, opened 2025-03-16
[BUG] deep_equals is broken for pandas.Index with None


XinyuWuu:
discovered in https://github.com/sktime/sktime/actions/runs/13868994536/job/38820293594?pr=7957

```py
from skbase.utils.deep_equals import deep_equals
import pandas

a = pandas.DataFrame([1, 2, 3])
b = pandas.DataFrame([1, 2, 3])
print(deep_equals(a, b))
a = pandas.Index([1])
b = pandas.Index([1])
print(deep_equals(a, b))
a = pandas.Index([None])
b = pandas.Index([None])
print(deep_equals(a, b))
```
```
True
True
False
```

XinyuWuu:
A temporary fix is added to sktime in PR https://github.com/sktime/sktime/pull/7957 via commit [deep_equals for pandas.Index](https://github.com/sktime/sktime/pull/7957/commits/d574d903df7cb52de3aca3babb430c6ad00ac3f1).

PR #407, opened 2025-03-16
[BUG] fix `deep_equals` for `pandas.Index`

XinyuWuu:
fix https://github.com/sktime/skbase/issues/406


fkiraly (approved):
hm, ok - good spot.

We should probably refactor the if/elses to "default plugins" at some point.

PR #422, opened 2025-06-04
Added preprocessing logic; because the 'all_objects' function current…

DebjyotiRay:
[ENH] Add filter_tags preprocessing to accept strings and lists in all_objects and get_package_metadata

#### Reference Issues
 Fixes #8210 from sktime

#### Changes:
This PR enhances the all_objects and get_package_metadata functions by adding preprocessing logic for the filter_tags parameter. Previously, these functions only handled filter_tags as a dict. Now they accept more user-friendly input formats:
 i) String input: `filter_tags="A"` → converts to `{"A": True}`
 ii) List/tuple input: `filter_tags=["A", "B"]` → converts to `{"A": True, "B": True}`
 iii) Dict input: Existing functionality preserved (dict is copied to prevent modification)
 iv) Error handling: Clear TypeError for invalid input types

- [x] Preprocessing logic: Verify the string/list-to-dict conversion works correctly
- [x] Backward compatibility: Ensure existing dict-based usage continues to work unchanged.

### DIFF
--- a/skbase/lookup/_lookup.py
+++ b/skbase/lookup/_lookup.py
@@ -190,34 +190,21 @@ def _filter_by_tags(obj, tag_filter=None, as_dataframe=True):
     if tag_filter is None:
         return True
 
-    type_msg = (
-        "filter_tags argument of all_objects must be "
-        "a dict with str or re.Pattern keys, "
-        "str, or iterable of str, "
-        "but found"
-    )
-
-    if not isinstance(tag_filter, (str, Iterable, dict)):
-        raise TypeError(f"{type_msg} type {type(tag_filter)}")
+    if not isinstance(tag_filter, dict):
+        raise TypeError(
+            "tag_filter argument must be a dict with str keys, "
+            f"but found type {type(tag_filter)}"
+        )
 
     if not hasattr(obj, "get_class_tag"):
         return False
 
-    # case: tag_filter is string
-    if isinstance(tag_filter, str):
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
Thanks!

Also, please do not change the logic of the existing tests.
Changing the error message to match the argument is of course ok.

DebjyotiRay:
Hello!
I tried to keep the logic of most of the existing tests intact.
Only changed some input and output formats of the tests corresponding to my changes, and added fresh ones for my edge cases.

Please let me know your review on them.

skbase/lookup/_lookup.py
```diff
@@ -190,34 +190,23 @@ def _filter_by_tags(obj, tag_filter=None, as_dataframe=True):
     if tag_filter is None:
         return True
 
-    type_msg = (
```
fkiraly, line 193: it looks like the function already allowed for iterable of str or str

DebjyotiRay:
Hello @fkiraly !
I updated the changes. Please lmk your thoughts.

skbase/lookup/_lookup.py
fkiraly: it seems like you are still changing the logic itself, making it less general, and removing the useful error message.

fkiraly (changes requested):
Thanks!

It seems like you are still changing the logic itself, making it less general, and removing the useful error message.

Is this really necessary? Please comment.

Issue #442, opened 2025-08-27
[BUG] `all_objects` does not retrieve classes inheriting from metaclasses on python 3.9 and 3.10


fkiraly:
It seems that `all_objects` does not retrieve classes inheriting from metaclasses on python 3.9 and 3.10

SimonBlanke:
This issue should be solved by #457

PR #457, opened 2025-10-13
 [BUG] Fix Metaclass lookup

SimonBlanke:
#### Reference Issues/PRs

#442

- I used `_get_members_uw` instead of `inspect.getmembers` to retrieve class.
- `inspect.unwrap()` does not work with MagicAttribute. It does not return the class itself.
- Verify inspect.isclass(MagicAttribute) returns True before attempting to unwrap
- Use klass.__name__ and klass.__module__ instead of trying to use the unwrapped version's attributes
- Catch exceptions and fall back to the original object

No

##### For all contributions

##### For code contributions

### DIFF
--- a/skbase/tests/conftest.py
+++ b/skbase/tests/conftest.py
@@ -3,7 +3,6 @@
 from typing import List
 
 from skbase.base import BaseEstimator, BaseObject
-from skbase.utils.dependencies import _check_python_version
 
 __all__: List[str] = [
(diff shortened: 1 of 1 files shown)


SimonBlanke:
> question: the current code looks simpler and is valid for python 3.11 and upwards. Should we put a reminder for October 2026 to revert to the original state? Or leave it like this?

I think we should leave it. If it works for 3.10 and above, then I would consider it more stable.

PR #466, opened 2025-11-06
[ENH] change `_MetaObjectMixin._check_objects` default for `attr_name` to auto-detect

### LINKED ISSUE #159: [ENH] change `_check_objects` default for `attr_name` to auto-detect
I think it would be nice to have the `_check_objects` default for `attr_name` to auto-detect from the tag, i.e. the default being `None`, which means, the value of the `named_object_parameters` tag.

SimonBlanke:
#### Reference Issues/PRs

Closes #159

Sets `attr_name` to `None` and auto detect from "named_object_parameters" of `attr_name` remains `None`

##### For all contributions

##### For code contributions


SimonBlanke:
added another test 'test_check_objects_attr_name_custom_tag'. It sets a custom tag via `set_tags`.

I just realized, that I used a private method `_check_objects` for the tests. Normally, that should be a no-go, but it seems the method is not used anywhere else in the package.

fkiraly (changes requested):
I think the coupling to `self` should happen at least one layer higher, i.e., not in the checker function, but where it is being called.

Otherwise we have intransparent coupling in a function that looks like a pure function but actually is not.

SimonBlanke:
> I think the coupling to `self` should happen at least one layer higher, i.e., not in the checker function, but where it is being called.
> 
> Otherwise we have intransparent coupling in a function that looks like a pure function but actually is not.

I understand we have some hidden coupling going on in `_check_objects`. The method signature does not reveal how it works. The tags are surprisingly read without the user knowing. From your explanation I infer the problem is, that some high level orchestration stuff (object stats, method calls) is mixed with lower level logic. It would help if we separate this.

I'll think about how to solve this.

SimonBlanke:
> are there any other tags that get read in there?

No, I only see `attr_name = self.get_tag("named_object_parameters")` here. For the decoupling I would like to separate the code into the following methods: 

```python
def _check_objects(self, objs, attr_name=None, ...):
    """Orchestration/high-level layer."""
    if attr_name is None:
        attr_name = self.get_tag("named_object_parameters")

    validated = self._validate_objects(objs, attr_name, ...)
    return self._coerce_to_named_object_tuples(validated, ...)

@staticmethod
def _validate_objects(objs, attr_name, cls_type=None, ...):
    # All validation logic
```

SimonBlanke:
okay there is a lot going on here (or maybe this is just in my head). First I added [this note](https://github.com/sktime/skbase/pull/466/files#diff-e6197ccf8ea00f7080233f04b26c12e30a6e790958541927d724179387010da5R390), because `_check_objects` does not really make sense to me as a private method, right now.

`_validate_objects` is a staticmethod now containing all the validation logic, `_check_objects` handles the rest. I also added a lot of comments. It feels necessary here.

SimonBlanke:
> There is a linter error in a file I did not change: "skbase/tests/test_base.py:389:5: B043 Do not call delattr with a constant attribute value, it is not any safer than normal property access."

fixed!

fkiraly (approved):
Looks good to me!

PR #467, opened 2025-11-08
[BUG] ensure `BaseMetaObject` resets at `set_params`; add 'test_set_params_resets_fitted_state' test to trigger error

### LINKED ISSUE #412: [BUG] `BaseMetaObject` does not `reset` at `set_params` call

SimonBlanke:
#### Reference Issues/PRs

fixes #412

This PR adds a test to trigger the bug described in #412. If desired this PR will also add a fix to the bug.

##### For all contributions

##### For code contributions

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -49,7 +49,7 @@ jobs:
           python -m build --wheel --sdist --outdir wheelhouse
 
       - name: Store wheels
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: wheels
           path: wheelhouse/*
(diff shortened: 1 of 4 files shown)


SimonBlanke:
One problem in #413 is, that reset is called multiple times in some instances. It would be better to always have one reset. I attempt to fix this problem and the original issue in this commit 07044cbb09f0f1dd77e0eaa7cbfffc1456268fd2

fkiraly (approved):
Brilliant!

I now understand that the `return self` call is the culprit.

This is unfortunate, I did spot it as an inconsistency and indeed it actually turned out (by your diagnosis) that it is the problem here.

If the clause `if params is None: return self` would not happen, would the issue not have happened? That is a strong indication that this special case caused a problem.

Either way, that is out of scope for this PR, but perhaps we should open an issue on whether this needs a change?

fkiraly:
issue opened here: https://github.com/sktime/skbase/issues/469

PR #476, opened 2025-11-22
[ENH] isolate lookup tests to dedicated mock package

### LINKED ISSUE #114: Reducing scope of `test_get_package_metadata_returns_expected_results`?
Adding a new function or class in most of `skbase`'s modules will trigger a failure of `test_get_package_metadata_returns_expected_results`, since `skbase` itself is being used as a test case for the lookup functionality, with expected retrieved functions and classes hard coded in the test config.

I don't think this is such a good idea, as the tests couple unrelated functionality to the `lookup` module.

E.g., it is not possible to refactor modules of `skbase` without triggering the test failure that is logically unrelated.

I would suggest to reduce the scope of `test_get_package_metadata_returns_expected_results`, possibly substantially, to the `mock_package` or that and a few other places?

arnavk23:
#### Reference Issues/PRs
Fixes #114.

This PR isolates `lookup` tests from the main package by introducing a dedicated test-only package, `skbase.tests.mock_package`.

The lookup tests scan this small controlled package instead of scanning `skbase` directly. This decouples lookup tests from unrelated production-package refactors and prevents brittle failures when internals outside lookup change.

No.

- Whether the new `mock_package` expectations accurately model intended `lookup` behavior.
- Whether test coverage still protects public and non-public module/item filtering semantics.

#### Any other comments?
The intent is to make lookup tests deterministic and focused on lookup behavior, not on unrelated evolution of the production package structure.

##### For all contributions

##### For code contributions


arnavk23:
@fkiraly please review this pr. Thanks!

fkiraly (changes requested):
Great!

There is though an entire config file which checks expectations against actuals after retrieval.

I would expect that still to happen for the mock package, but it does not - are you sure you have moved all tests over? I would at least expect the "full package expected lookup results" to be removed.

arnavk23:
> I would at least expect the "full package expected lookup results" to be removed.

I've removed the "full package expected lookup results" from `conftest.py` eliminating the SKBASE_*_BY_MODULE dictionaries that contained the expected classes and functions for the entire skbase package. I also removed the test function test_get_package_metadata_returns_expected_results that relied on those expected results, as it was checking the full package lookup against predefined expectations.

The checks for the mock package still occur through the existing tests:
- `test_all_objects_returns_expected_output` verifies that `all_objects` returns the expected classes from `MOCK_PACKAGE_OBJECTS` for the mock package.
- `test_get_package_metadata_matches_expected_representation_mock_package` builds expected metadata by introspecting the mock package modules and compares it against the actual results from `get_package_metadata`.

This ensures that expectation checks against actual retrieval still happen for the mock package, while restricting the lookup tests by removing the broader full-package validation.

fkiraly (changes requested):
It feels like you are using AI to make changes.

You have made some changes related to my review request, but `_conftest` looks like a combination of changes that are sensible, some that are random and introduce issues.

Can you comment - after thinking about what you are doing? How did you address the changes and why?

fkiraly:
Also, the mock package does not have private modules or functions - this reduces test coverage.

Coverage should not be reduced.

arnavk23:
@fkiraly Thank you for pointing that out. I have made some changes. Please review and do share if there is anything you need.

PR #511, opened 2026-03-03
[BUG] Fix incorrect tag resolution in `_get_class_flags` with diamond inheritance

WHOIM1205:
### Summary

`_FlagManager._get_class_flags` currently uses `hasattr` and `getattr` when iterating over the class MRO.  
Because `getattr` follows Python's full attribute lookup chain, classes that do not define `_tags` (or `_config`) themselves may still return values inherited from ancestors.

In diamond inheritance scenarios this can re-apply ancestor flags later in the merge order and overwrite overrides coming from mixins.

### Example

```python
class A(BaseObject):
    _tags = {"handles_missing_data": False}

class MissingDataMixin(A):
    _tags = {"handles_missing_data": True}

class MyBase(A):
    pass

class Concrete(MyBase, MissingDataMixin):
    _tags = {"custom_flag": True}

Concrete().get_tag("handles_missing_data")


fkiraly (changes requested):
Good catch. I assume an AI caught this.

Can you please make sure to:
* remove the AI generated comment and leave the old comment there
* fix the issue title

fkiraly (changes requested):
Please add a test.

fkiraly (approved):
Thanks

PR #512, opened 2026-03-04
[ENH] Add regression test ensuring filter_tags in all_objects does not return empty results

### LINKED ISSUE #162: [ENH] ensure test coverage of #161 failure case
We should ensure test coverage of the failure case in https://github.com/sktime/skbase/pull/161

shourya1419-netizen:
Summary

Adds a regression test to ensure that filter_tags in all_objects does not return empty results when matching objects exist.

What this PR does

Adds a test case verifying that objects with matching tags are correctly returned.

Prevents regression of the bug described in #162 where all_objects(filter_tags=...) returned an empty result.

Related Issue

Closes #162


fkiraly (changes requested):
Why is this almost identical to changes in https://github.com/sktime/skbase/pull/476?

fkiraly (changes requested):
I see. Please do that without writing code at runtime.

fkiraly (approved):
Looks good to me now, thanks!

PR #514, opened 2026-03-04
Docs add docstring examples

### LINKED ISSUE #19: [DOC] Add examples to docstrings

shourya1419-netizen:
Adds usage examples to several `BaseObject` related docstrings in `skbase/base/_base.py`.

These examples demonstrate basic usage such as creating a `BaseObject`, configuring it with `set_config`, and retrieving configuration using `get_config`. The goal is to improve documentation clarity and help new users understand how to use these methods in practice.

Closes #19

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -73,6 +73,13 @@ class BaseObject(_FlagManager):
     """Base class for parametric objects with sktime style tag interface.
 
     Extends scikit-learn's BaseEstimator to include sktime style interface for tags.
+
+    Examples
+    --------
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Adding docstrings to these methods would be confusing, since these are methods that get inherited throughout the ecosystem, present in objects that are not `BaseObject`, and not visibly related to it by the user.

I will hence close the PR.

Could you add examples in other places instead?

PR #519, opened 2026-03-07
[BUG] Fix deep_equals failing on string-dtype numpy arrays (numpy < 2.0)

### LINKED ISSUE #517: [BUG] `deep_equals` fails on string-dtype numpy arrays (numpy < 2.0)
**Describe the bug**

Calling `deep_equals` (or anything that triggers it, like `clone` + `set_params` verification)
on a numpy array with string dtype crashes with:

```
TypeError: ufunc 'isnan' not supported for the input types,
and the inputs could not be safely coerced to any supported types
according to the casting rule 'safe'
```
It affects any `BaseObject` with a numpy string array parameter, when using numpy < 2.0

**To Reproduce**

```python
import numpy as np
from skbase.utils.deep_equals import deep_equals

s = np.array(["1", "2", "1"])
deep_equals(s, s)
```

This raises a TypeError

**Expected behavior**

No TypeError

**Environment**

The bug exists in scikit-base 0.13.1 with numpy < 2.0. I tested it on ubuntu, but this should not affect the bug.

**Additional context**

shourya1419-netizen:
Fixes #517

## What was the bug?
`deep_equals` crashed with a `TypeError` when comparing numpy arrays 
with string dtype on numpy < 2.0.

The check `if x.dtype == "str"` in `_numpy_equals_plugin` never matched 
in numpy < 2.0, because string arrays have dtype like `U3` or `<U1` 
(not `"str"`). This caused string arrays to fall through to the `else` 
branch which calls `np.array_equal(x, y, equal_nan=True)` — and 
`equal_nan=True` internally calls `np.isnan()`, which does not support 
string types.

## Fix
Added `np.issubdtype(x.dtype, np.character)` to the dtype check, which 
correctly catches all string dtypes across numpy versions (both < 2.0 
and >= 2.0).

## Testing
Added a new test file `tests/test_deep_equals.py` covering:
- equal string arrays return `True`
- unequal string arrays return `False`

All tests pass.


fkiraly (changes requested):
Thanks! The test would not catch it though because afaik we are testing only with `numpy 2`. I also opened an issue, hence, to add test coverage here: #532

fkiraly (approved):

fkiraly:
Abandoned PR. Merging without added test since the fix seems reasonable.

PR #537, opened 2026-03-24
[MNT] downwards compatibility testing for deep_equals

sid200727:
Adds a new GitHub Actions workflow `.github/workflows/test_deep_equals_compat.yml`
that tests `deep_equals` against older versions of numpy and pandas using uv's
`--exclude-newer` flag, as proposed in #532.

Three matrix jobs are added:
- `numpy-1x-cutoff` — cutoff Jan 2024, resolves numpy 1.x (before numpy 2.0 release)
- `pandas-1x-cutoff` — cutoff Jan 2023, resolves pandas 1.x era
- `pandas-2x-cutoff` — cutoff Jun 2024, resolves pandas 2.x era

Only `skbase/utils/tests/test_deep_equals.py` is run in each job, since
`deep_equals` is the only utility in skbase that has special-case handling
for numpy and pandas objects.
No new dependencies. The workflow uses uv (astral-sh/setup-uv@v4) which is
already available as a GitHub Actions action and is not a package dependency
of skbase itself.
- The choice of cutoff dates for each matrix job (Jan 2024, Jan 2023, Jun 2024).
- Whether the install command `uv pip install --system --exclude-newer <date> -e ".[all_extras,dev]" numpy pandas` correctly resolves the intended older versions.
The approach (uv --exclude-newer to simulate older environments) was agreed
with @fkiraly in #532. Happy to adjust cutoff dates or add further timepoints
if needed.


fkiraly (changes requested):
Thanks!

* could we simply run that on each PR instead of in a CRON? Seems like a small enough test.
* can you use the same "display dependencies" step as in the remainder of the CI, instead of print statements?

fkiraly:
still a CRON

sid200727:
Hi @fkiraly, I've fixed the pull_request trigger and it should now run on each PR as requested, along with the CRON schedule. Could you take another look? Thank you!

.github/workflows/test_deep_equals_compat.yml
```diff
@@ -0,0 +1,56 @@
+# copyright: skbase developers, BSD-3-Clause License (see LICENSE file)
+name: Test deep_equals downwards compatibility
+
+on:
+  pull_request:
+  schedule:
```
fkiraly: this is still a CRON job

fkiraly (approved):
I fixed the remaining requests and moved the job to the primary set of tests.

sid200727:
Thank you for fixing the remaining issues and merging @fkiraly ! I learned a lot from this contribution. 🙏

PR #540, opened 2026-04-04
[BUG] Fix _get_class_flags dropping Mixin flags under multiple inheritance

### LINKED ISSUE #539: [BUG] _get_class_flags dynamically slices the MRO index, permanently dropping tags for Mixins during multiple inheritance
**Describe the bug**
The core component `_get_class_flags` iterates over `reversed(inspect.getmro(cls)[:-2])`. The arbitrary `[:-2]` cut off removes the last two members of the object tree. In multiple-inheritance situations (e.g., `class CustomAlgorithm(BaseObject, ArbitraryMixin)`), `ArbitraryMixin` slides against the end index of the MRO. Consequently, its associated `_tags` and `_flags` properties get unconditionally deleted out of `get_class_flags`.

**To Reproduce**
```python
import inspect
from skbase.base import BaseObject

class MyMixin:
    _tags = {"mixin_tag": "should_be_seen"}

# With multiple inheritance
class MyObj(BaseObject, MyMixin):
    _tags = {"obj_tag": "seen"}

tags = MyObj.get_class_tags()
print("Tags:", tags) # mixin_tag will be missing completely from the output due to the [:-2] slice logic
```

**Expected behavior**
Tags defined in all class ancestor mixins should be exposed regardless of multiple inheritance order or length (except for Python's base `object` which never contains flags).

**Proposed Fix**
Revert to iterating through the `getmro()` resolution tree entirely natively while explicitly discarding `object` rather than generic indices:
```python
for parent_class in reversed(inspect.getmro(cls)):
    if parent_class is object:
        continue
    # ... check hasattr
```

vedant21-oss:
## Summary

Fixes #539 - _FlagManager._get_class_flags dropped flags from user-defined Mixin classes under multiple inheritance due to a hardcoded [:-2] MRO slice.

## Root Cause

The previous code assumed the last two MRO entries are always BaseObject and object:

```python
for parent_class in reversed(inspect.getmro(cls)[:-2]):
```

Under Python's C3 linearisation, class MyObj(BaseObject, MyMixin) produces:

```
MyObj -> BaseObject -> _FlagManager -> MyMixin -> object
```

MyMixin sits at position [-2], so the slice permanently drops it and all its flags - silently, with no error.

## Fix

Iterate the full MRO and skip only object via an explicit identity check:

```python
for parent_class in reversed(inspect.getmro(cls)):
    if parent_class is object:
        continue
```

object is always the last entry and can never carry user-defined flags, making it the only safe sentinel to exclude.

## Changes

- skbase/base/_tagmanager.py: Replace [:-2] slice with if parent_class is object: continue. Add detailed comment explaining the root cause so this is never accidentally reverted.
- - skbase/tests/test_meta.py: Add missing _FlagManager import + add test_get_class_flags_preserves_mixin_tags regression test that pins this fix and will fail loudly if the slice returns.
## Testing

```
11 passed in 0.02s
```


skbase/base/_tagmanager.py
```diff
@@ -32,9 +32,24 @@ class attribute via nested inheritance. NOT overridden by dynamic
...
+        # Python's C3 linearisation can push user-defined Mixin classes to the
+        # very end of the MRO (just before `object`). For example:
+        #
+        #   class MyObj(BaseObject, MyMixin): ...
+        #   # MRO: MyObj -> BaseObject -> _FlagManager -> MyMixin -> object
+        #
+        # Slicing with [:-2] drops `MyMixin`, permanently losing any flags it
+        # defines. The only safe sentinel is `object` itself (which never carries
+        # user-defined flags), so we skip it explicitly and iterate the rest.
+        for parent_class in reversed(inspect.getmro(cls)):
```
fkiraly, line 37: if we want to make this general, do we not want to drop it entirely?

Also, please do not add comments that explain the previous situation, they do not make sense to a reader of the code base. AIs often add comments that reference a change.

fkiraly (approved):
I fixed it.

PR #553, opened 2026-05-01
[BUG] Fix in-place mutation in make_strings_unique

### LINKED ISSUE #551: make_strings_unique mutates the caller's input list in-place instead of returning a modified copy
## Summary
`make_strings_unique(str_list)` assigns `unique_strs = str_list` (an alias,
not a copy) and then modifies `unique_strs[i]`, which silently mutates
the original `str_list` the caller passed in. Any code that retains a
reference to the original list will observe unexpected changes.

## Reproduction
```python
from skbase.utils._iter import make_strings_unique
names = ["a", "a", "b"]
result = make_strings_unique(names)
print(names)   # ['a_1', 'a_2', 'b'] -- original was mutated!
```

## Expected Behavior
The input should not be modified. Only the returned list should be unique.

## Suggested Fix
Replace `unique_strs = str_list` with `unique_strs = list(str_list)`.

SHIVANSH-ux-ys:
Fixes #551

The function `make_strings_unique` was mutating the input list in-place due to a direct reference assignment. This PR ensures a copy is created before modification.

Verified with a reproduction script:
```python
mylist = ["a", "b", "a"]
make_strings_unique(mylist)
# mylist remains ["a", "b", "a"]
```

### DIFF
--- a/skbase/utils/_iter.py
+++ b/skbase/utils/_iter.py
@@ -210,7 +210,7 @@ def make_strings_unique(str_list):
     str_count = collections.Counter(str_list)
     # if any duplicates, we append _integer of occurrence to non-uniques
     now_count: collections.Counter = collections.Counter()
-    unique_strs = str_list
+    unique_strs = list(str_list)
     for i, x in enumerate(unique_strs):
         if str_count[x] > 1:
             now_count.update([x])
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
Great, thanks!

Can you add a test against regression? I.e., testing that the function no longer mutates its input?

SHIVANSH-ux-ys:
Hi @fkiraly,

I have added a regression test in `skbase/utils/tests/test_iter.py` to ensure that `make_strings_unique` does not mutate its input lists (verifying both flat and nested list inputs). 

The test passes successfully. Ready for another review!

fkiraly (approved):
Thanks!

PR #555, opened 2026-05-02
[ENH] respect explicit default in get_tag to match getattr semantics

### LINKED ISSUE #291: Behaviour of `get_tag` is different from standard expectation from `get` or `getattr`
`get_tag` is behaving differently from `get` or `getattr` in python. Usually they return default value provided if key is missing. In `skbase`, it fails even when default value is provided, unless one specifically pass an additional argument to control error behaviour.

This is not really a bug, but a behaviour that is probably not very intuitive to python users.

```pycon
>>>
>>> from skbase.base import BaseObject
>>>
>>> BaseObject().get_tag("abcd")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/anirban/conda-environments/sktime/lib/python3.10/site-packages/skbase/base/_base.py", line 578, in get_tag
    return self._get_flag(
  File "/home/anirban/conda-environments/sktime/lib/python3.10/site-packages/skbase/base/_tagmanager.py", line 147, in _get_flag
    raise ValueError(f"Tag with name {flag_name} could not be found.")
ValueError: Tag with name abcd could not be found.
>>>
>>> BaseObject().get_tag("abcd", tag_value_default="pqrs")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/anirban/conda-environments/sktime/lib/python3.10/site-packages/skbase/base/_base.py", line 578, in get_tag
    return self._get_flag(
  File "/home/anirban/conda-environments/sktime/lib/python3.10/site-packages/skbase/base/_tagmanager.py", line 147, in _get_flag
    raise ValueError(f"Tag with name {flag_name} could not be found.")
ValueError: Tag with name abcd could not be found.
>>>
>>> ARIMA().get_tag("abcd", tag_value_default="pqrs", raise_error=False)
'pqrs'
>>>
```

Deep-Axe:
#### Reference Issues/PRs

Fixes #291 

This PR modifies ```get_tag``` (and the internal ```_get_flag```) to respect an explicitly provided ```tag_value_default```. Currently, ```get_tag``` raises a ```ValueError``` if a tag is missing even if a default value is supplied, unless ```raise_error=False``` is also explicitly passed. This behavior is unintuitive, and inconsistent with ```get_class_tag``` and standard patterns like ```getattr()``` or ```dict.get()```

***Changes:***
- Introduced a ```_DEFAULT_SENTINEL``` in ```_tagmanager.py``` to reliably distinguish between an omitted default and an explicit None default.
- Updated ```_get_flag``` to bypass the ```raise_error``` check if a default was provided.
- Updated public method signatures and docstrings in ```_base.py``` and ```_tagmanager.py``` to clarify the new behavior.
- Added unit tests in ```test_base.py``` to verify that explicit defaults (including None) correctly suppress the ValueError without needing raise_error=False.

***Backward Compatibility:***
The change is fully backward compatible. Calling ```get_tag("missing")``` without a second argument will still raise a ValueError by default, preserving current error-checking behavior for typos.

No

 - The use of the ```_DEFAULT_SENTINEL``` object to ensure ```get_tag("name", None)``` correctly returns None instead of raising an error.
 - The alignment of the ```get_tag``` logic with the existing behavior of ```get_class_tag.```

#### Any other comments?
As discussed in the issue thread, a follow-up PR could be opened later to start a ```FutureWarning``` deprecation cycle for changing the global default of ```raise_error``` from True to False, but this PR focuses on respecting the user-supplied default first.

##### For all contributions
- [x] The PR title starts with either [ENH], [CI/CD], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, CI/CD, maintenance, documentation, or a bug.

##### For code contributions

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -181,6 +181,16 @@
       "contributions": [
         "infra"
       ]
+    },
+    {
+      "login": "Deep-Axe",
+      "name": "Deep-Axe",
+      "avatar_url": "https://avatars.githubusercontent.com/u/152912444?v=4",
+      "profile": "https://github.com/Deep-Axe",
+      "contributions": [
+        "code",
+        "test"
+      ]
     }
   ],
   "projectName": "skbase",
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -63,9 +63,9 @@ from typing import List
 from skbase._exceptions import NotFittedError
 from skbase.base._clone_base import _check_clone, _clone
 from skbase.base._pretty_printing._object_html_repr import _object_html_repr
-from skbase.base._tagmanager import _FlagManager
+from skbase.base._tagmanager import _DEFAULT_SENTINEL, _FlagManager
 
-__author__: List[str] = ["fkiraly", "mloning", "RNKuhns", "tpvasconcelos"]
+__author__: List[str] = ["fkiraly", "mloning", "RNKuhns", "tpvasconcelos", "Deep-Axe"]
 __all__: List[str] = ["BaseEstimator", "BaseObject"]
 
 
@@ -601,7 +601,9 @@ class BaseObject(_FlagManager):
         """
         return self._get_flags(flag_attr_name="_tags")
 
-    def get_tag(self, tag_name, tag_value_default=None, raise_error=True):
+    def get_tag(
+        self, tag_name, tag_value_default=_DEFAULT_SENTINEL, raise_error=True
+    ):
         """Get tag value from instance, with tag level inheritance and overrides.
 
         Every ``scikit-base`` compatible object has a dictionary of tags.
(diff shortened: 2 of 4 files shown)


fkiraly (changes requested):
Hm, can you kindly explain why you think this change is needed, or why it would be an improvement? In which usage scenario, etc.

I also want to point out that this is *not* backwards compatible and in a quite central place in the API.
Namely, in case that `raise_error=True` and passing `tag_value_default`, the behaviour changes.

Deep-Axe:
Hi @fkiraly
The motivation was from the thread in Issue #291. 
```get_tag("name","fallback")``` raises a ```ValueError``` even though the caller explicitly provided a fallback.
This defeats the purpose of ```tag_value_default``` unless the user also knows to pass ```raise_error=False```. This deviates from standard Python patterns (getattr, dict.get) and is
inconsistent with ```get_class_tag```, which already returns the supplied default unconditionally.

The PR does change behavior for the case where a caller passes ```tag_value_default``` but leaves ```raise_error``` as its default True.
While this call pattern is contradictory (saying "use this fallback if missing" and "raise an error if missing" simultaneously), but it shouldn't silently change.

As you suggested in the original issue thread, a deprecation cycle would be a place to start.
Would you like me to update this PR so that passing a default while ```raise_error=True``` emits a ```FutureWarning``` (warning of the future getattr-like behavior) but still raises the ValueError for now?

fkiraly:
Hm, having thought about this: I think we should not change the default behaviour of `get_tag`.

The reason for this is that:

* per design expectation, default tag values should be set in a common super class, and not in the `get_tag` method.
* more generally, default values for tag should be managed *globally* (inside a package) and not *locally* (at call)
* in most downstream uses which do follow this design, it feels preferable to have a failure if the tag is not actually present, but requested

I will add these thoughts to the original issue.

PR #559, opened 2026-06-02
[BUG] Fix get_params(deep=True) for BaseObject class parameters

### LINKED ISSUE #558: [BUG] get_params(deep=True) fails when an __init__ parameter holds a BaseObject subclass (type), not an instance
**Describe the bug**

BaseObject.get_params(deep=True) treats any __init__ parameter whose value has a get_params attribute as a nested component and calls value.get_params() on it. That fails when the parameter holds a class (a BaseObject subclass), not an instance: scikit-base ends up calling SomeClass.get_params() without self, which raises TypeError: BaseObject.get_params() missing 1 required positional argument: 'self'.

The same logic affects the HTML diagram repr (_object_html_repr -> _get_visual_block), which calls get_params() with default deep=True.

This was hit in [sktime/skpro#924](https://github.com/sktime/skpro/pull/924): MOMFitter stores a distribution class (e.g. Normal) in dist_cls. Without local overrides to get_params and _repr_html_, introspection and Jupyter diagram display fail.

**To Reproduce**

```
from skbase import BaseObject

class Child(BaseObject):
    def __init__(self, x=1):
        self.x = x
        super().__init__()

class Parent(BaseObject):
    def __init__(self, nested_cls):
        self.nested_cls = nested_cls  # class, not instance
        super().__init__()

p = Parent(nested_cls=Child)
p.get_params(deep=True)

```

**Expected behavior**

get_params(deep=True) should return top-level parameters (e.g. nested_cls=<class 'Child'>) without calling get_params() on Child the class.

Nested parameters should only be expanded when the value is a BaseObject instance, not a type. Plain repr() using get_params(deep=False) can continue to work as today.

**Environment**

**Additional context**

Suggested fix: use isinstance(value, BaseObject) (or equivalent) instead of hasattr(value, "get_params") when descending in get_params(deep=True) and in _get_visual_block.

sklearn-style estimators generally only recurse into nested instances; aligning scikit-base would avoid workarounds in downstream packages (e.g. skpro MOMFitter).

patelchaitany:
## Summary

Fixes #558.

`get_params(deep=True)` and the HTML diagram repr (`_get_visual_block`) previously treated any parameter value with a `get_params` attribute as a nested component. That breaks when the parameter holds a **BaseObject subclass (type)** rather than an instance—for example `dist_cls=Normal` in skpro's `MOMFitter`-because `SomeClass.get_params()` is invoked without `self`.

This PR aligns recursion with scikit-learn-style behavior: only descend into **BaseObject instances**, using `isinstance(value, BaseObject)` instead of `hasattr(value, "get_params")`.

## Changes

- `BaseObject.get_params`: recurse only for `isinstance(value, BaseObject)`
- `_get_visual_block`: same check for first-layer components (lazy import of `BaseObject` to avoid circular imports)
- Regression tests for `get_params(deep=True)` and HTML repr with a class parameter

## Test plan

- [x] `pytest skbase/tests/test_base.py::test_get_params_deep_with_baseobject_class_param`
- [x] `pytest skbase/base/_pretty_printing/tests/test_object_html_repr.py::test_html_repr_with_baseobject_class_param`
- [x] `pytest skbase/tests/test_base.py -k "get_params or repr_html"`

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -58,6 +58,7 @@ import re
 import warnings
 from collections import defaultdict
 from copy import deepcopy
+from inspect import isclass
 from typing import List
 
 from skbase._exceptions import NotFittedError
@@ -335,7 +336,7 @@ class BaseObject(_FlagManager):
             for key, value in params.items():
                 # Skip class objects (e.g. BaseObject subclasses stored as params);
                 # only recurse into instances with a parameter interface (#558).
-                if hasattr(value, "get_params") and not isinstance(value, type):
+                if hasattr(value, "get_params") and not isclass(value):
                     deep_items = value.get_params().items()
                     deep_params.update({f"{key}__{k}": val for k, val in deep_items})
             params.update(deep_params)
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -333,7 +333,9 @@ class BaseObject(_FlagManager):
         if deep:
             deep_params = {}
             for key, value in params.items():
-                if hasattr(value, "get_params"):
+                # Skip class objects (e.g. BaseObject subclasses stored as params);
+                # only recurse into instances with a parameter interface (#558).
+                if hasattr(value, "get_params") and not isinstance(value, type):
                     deep_items = value.get_params().items()
                     deep_params.update({f"{key}__{k}": val for k, val in deep_items})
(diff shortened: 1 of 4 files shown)


fkiraly (approved):
This looks like a fix, thanks! Very nice!

PR #574, opened 2026-07-18
&toller892 [ENH] add tests for use of `tag_value_default` in `get_tag` and `get_class_tag`

fkiraly:
adds test coverage for use of `tag_value_default` in `get_tag` and `get_class_tag`, following tests in https://github.com/sktime/sktime/pull/10312 by @toller892


fkiraly:
looking at https://github.com/sktime/sktime/pull/10312 again (and the copilot comments) again, it appears that the PR is based on a misunderstanding of the docstring and/or overuse of AI.

The defaulting behaviour is already tested in `test_get_class_tag` and `test_get_tag`, and does not need additional tests.

PR #586, opened 2026-08-24
[BUG] mute stderr in QuickTester.run_tests, and fix np.all generator checks

yash-sangwan:
#### Reference Issues/PRs

Spun out of the `sktime` test framework refactor (#10647), specifically PR #10862 which inherits `BaseFixtureGenerator` and `QuickTester` from `skbase.testing`. @fkiraly suggested upstreaming the `StderrMute` change there.

This introduces two independent fixes in `skbase/testing/test_all_objects.py`, both found while making `sktime` inherit this test framework.

**1. `run_tests` only muted stdout**

Currently, `run_tests` wraps each test in `StdoutMute(active=verbose < 2)`, so anything a test writes to stderr still reaches the log at `verbose=0`. `sktime` has been muting both streams since PR #8799, so inheriting `run_tests` unchanged would have silently dropped the stderr half. `StderrMute` already shipped in `skbase.utils.stderr_mute`, it was just never used here.

Measured before and after:

| | stdout leak | stderr leak |
| --- | --- | --- |
| before, `verbose=0` | 0 | 1 |
| after, `verbose=0` | 0 | 0 |
| after, `verbose=2` | 1 | 1 |

`verbose=2` still shows both streams, so the escape hatch is unchanged.

**2. Three checks passed a generator expression to `np.all`**

For example:
```python
if not np.all(isinstance(x, str) for x in obj):

```

Without brackets, this is a generator expression. `np.all` receives a single generator object rather than an array of booleans. NumPy treats it as one scalar object and applies truthiness, and every generator object is truthy. The result is unconditionally `True`, the generator is never consumed, and the check never fires.

Affected areas:

* `QuickTester._check_none_str_or_list_of_str`: `tests_to_run`, `fixtures_to_run`, and related args were effectively unvalidated.
* Both checks in `TestAllObjects.test_create_test_instances_and_names`.

There is a secondary bug in the last of these. It read:

```python
assert np.all(isinstance(name, names) for name in names), "... all must be strings"

```

Since `names` is a list of strings, not a type, `isinstance(name, names)` is invalid and raises a `TypeError`. The generator bug was masking it. Adding brackets alone would have converted a silently passing assert into a hard `TypeError` on every object under test, so this is corrected to `isinstance(name, str)` to match the assertion message.

No

* Whether muting stderr by default at `verbose < 2` is the desired behavior, or whether it should be opt-in.
* The `names` to `str` correction in `test_create_test_instances_and_names`. It changes what that assert actually checks since it has never truly executed before.

### DIFF
--- a/skbase/testing/test_all_objects.py
+++ b/skbase/testing/test_all_objects.py
@@ -342,6 +342,7 @@ class QuickTester:
         """
         from _pytest.outcomes import Skipped
 
+        from skbase.utils.stderr_mute import StderrMute
         from skbase.utils.stdout_mute import StdoutMute
 
         tests_to_run = self._check_none_str_or_list_of_str(
@@ -464,7 +465,10 @@ class QuickTester:
                 print_if_verbose(f"{key}")
 
                 try:
-                    with StdoutMute(active=verbose < 2):
+                    with (
+                        StdoutMute(active=verbose < 2),
+                        StderrMute(active=verbose < 2),
+                    ):
                         test_fun(**deepcopy(args))
                     results[key] = "PASSED"
                     print_if_verbose("PASSED")
@@ -541,7 +545,7 @@ class QuickTester:
                 obj = [obj]
             if not isinstance(obj, list):
                 raise ValueError(msg)
-            if not np.all(isinstance(x, str) for x in obj):
+            if not np.all([isinstance(x, str) for x in obj]):
                 raise ValueError(msg)
         return obj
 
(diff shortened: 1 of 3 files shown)


fkiraly (changes requested):
Thanks, and also for fixing the bugs in `np.all`.

Could you kindly add tests to prevent regressions? There should be fixtures that you can adapt. If it is too complex (e.g., resulting in 100s lines of code) we can omit it though.

yash-sangwan:
Just added! I managed to keep it very concise (under 50 lines) by using `pytest`'s native `capsys` fixture to capture the stream outputs, so no complex mocking was needed.
I added regression tests for the `StderrMute` behavior and the `QuickTester._check_none_str_or_list_of_str` input validation. I omitted the `test_create_test_instances_and_names` dummy objects to respect your note about keeping the footprint small.
Lmk if this looks good to go.

fkiraly (approved):
Perfect.

PR #589, opened 2026-08-28
[BUG] roll back state when set_params fails validation

### LINKED ISSUE #10695: [BUG] failed `set_params` leaves estimator in a state `__init__` cannot produce
## LLM generated content, by Claude Opus 5

Drafted with AI assistance and reviewed by me; I reproduced the behaviour locally before filing.

**Describe the bug**

If `set_params` is called with a value that parameter validation rejects, the estimator is left holding the rejected value. Validation in `__init__` (or `__post_init__`) runs *after* the parameter is written to `self`, so the raise does not undo the write. `get_params` then returns the invalid value, and `clone` and `fit` fail from that point on.

**To Reproduce**

```python
from sklearn.base import clone

from sktime.transformations.dilation_mapping import DilationMappingTransformer

d = DilationMappingTransformer(dilation=2)

try:
    d.set_params(dilation=0)
except ValueError:
    pass

print(d.dilation)        # 0  - the rejected value was written anyway
print(d.get_params())    # {'dilation': 0}

clone(d)                 # ValueError: Dilation must be greater than 0
```

`DilationMappingTransformer(dilation=0)` raises and produces no object, so `dilation=0` is a state `__init__` cannot produce - but `set_params` produces it.

**Expected behavior**

A failed `set_params` should leave the estimator in the state it was in before the call.

This is the requirement adopted in #2573, where the conclusion was that `"__init__` has no content" is an implementer-facing artifact of `scikit-learn`, and that the important, user-facing requirement is that `set_params` leaves the same state as `__init__` would.

**Additional context**

Not specific to `DilationMappingTransformer` - `ADICVTransformer` reproduces the same way with `features`. On current `main` about 110 classes raise inside `__init__` or `__post_init__`, so the pattern is widespread, though I have not checked how many of those write before raising.

Verified unchanged on `main` at v1.1.0 - `dilation_mapping.py` still assigns `self.dilation` before the `raise`; the versions block below is from a 1.0.1 editable install.

`test_set_params_sklearn` cannot catch this: it states that it uses the other test parameter settings "which are assumed valid ... to play along with the `__init__` content", so no invalid value is ever passed to `set_params` in the suite.

#910 fixed a mutation variant of the same class of problem for `Prophet` in 2021, where `__init__` coerced values rather than raising.

There seem to be a few possible directions - validating before writing to `self`, making `set_params` restore the previous parameters if the reconstruction raises, or adding a framework test that asserts the pre-call state survives a rejected `set_params`. Which of those fits the intended design best, and is this worth fixing?

**Versions**
<details>

```
machine: Linux-7.0.12-zen1-1-zen-x86_64-with-glibc2.43

Python dependencies:
          pip: 26.1.2
       sktime: 1.0.1
 scikit-learn: 1.7.2
  scikit-base: 1.0.2
        numpy: 2.4.6
        scipy: 1.18.0
       pandas: 2.3.3
       joblib: 1.5.3
        skpro: 2.14.0
```

</details>

OfficialAbhinavSingh:
#### Reference Issues/PRs

Fixes https://github.com/sktime/sktime/issues/10695. See also https://github.com/sktime/sktime/pull/10821, an earlier attempt at this issue that patched `set_params` on the `sktime` side and was closed unmerged, since the bug is in `BaseObject.set_params` here in `skbase`, not in `sktime`.

`BaseObject.set_params` writes new parameter values to `self` via `setattr` before calling `self.reset()`, which is what actually re-runs `__init__` to validate them. If `__init__` rejects a value, the raise happens after the write, so the rejected value survives on `self` -- a state `__init__` could never have produced. `get_params`, `clone`, and any later `set_params` call then operate on that invalid state.

```python
class Validated(BaseObject):
    def __init__(self, x=1):
        self.x = x
        if x < 0:
            raise ValueError("x must be >= 0")
        super().__init__()

v = Validated(x=5)
v.set_params(x=-1)   # raises ValueError
v.x                   # -1, before this fix -- the rejected value stuck
```

Fix snapshots `self.__dict__` before the parameter-writing loop in `set_params`, and restores it if anything in that loop or the following `self.reset()` call raises, then re-raises the original exception. Confirmed against the estimator that surfaced this in `sktime` (`DilationMappingTransformer`): `set_params(dilation=0)` now leaves `dilation` at its prior valid value instead of `0`, and `clone()` succeeds afterward.

No.

- Whether snapshotting/restoring `self.__dict__` is the right mechanism, versus validating before writing to `self` in the first place -- the issue write-up considered both directions and this seemed the smaller, more general change, since it doesn't require touching every `__init__`.
- The existing `test_get_params_after_set_params` test in `test_base.py` already asserts rollback semantics under fuzz values, but the fixture class it uses (`Parent`) never raises in `__init__`, so that assertion was never actually exercised. Added a new test with a fixture that does validate, `test_set_params_rolls_back_state_on_invalid_value`.

#### Any other comments?

An LLM was used as a search and navigation tool to help trace the bug to its root cause in `set_params`/`reset()` and to draft this description. Every claim above -- the repro, the root cause, the before/after behavior, and the test counts -- was produced by running the code locally on both sides of the fix, including a run of the new test against the pre-fix source to confirm it fails without the change.

##### For all contributions

##### For code contributions

### DIFF
--- a/skbase/base/_base.py
+++ b/skbase/base/_base.py
@@ -384,40 +384,48 @@ class BaseObject(_FlagManager):
             return self
         valid_params = self.get_params(deep=True)
 
-        # snapshot instance state, to restore it if setting params below raises,
-        # e.g., through a validation failure in __init__ during the reset call.
-        # Without this, a failed set_params can leave self in a state that
-        # __init__ could not have produced, since setattr writes below happen
-        # before that validation runs.
+        # snapshot instance state, to restore it if the reset call below raises,
+        # e.g., through a parameter validation failure in __init__.
+        # Must be taken before the setattr writes below, since those writes are
+        # what leave self in a state that __init__ could not have produced.
+        # This is a shallow copy of the instance __dict__, i.e., it stores
+        # references to attribute values, it does not copy the values themselves.
         prev_state = self.__dict__.copy()
 
         unmatched_keys = []
 
         nested_params = defaultdict(dict)  # grouped by prefix
-        try:
-            for full_key, value in params.items():
-                # split full_key by first occurrence of __, if contains __
-                # "key_without_dblunderscore" -> "key_without_dbl_underscore", None, None
-                # "key__with__dblunderscore" -> "key", "__", "with__dblunderscore"
-                key, delim, sub_key = full_key.partition("__")
-                # if key not recognized, remember for suffix matching
-                if key not in valid_params:
-                    unmatched_keys += [key]
-                # if full_key contained __, collect suffix for component set_params
-                elif delim:
-                    nested_params[key][sub_key] = value
-                # if key is found and did not contain __, set self.key to the value
-                else:
-                    setattr(self, key, value)
-                    valid_params[key] = value
+        for full_key, value in params.items():
+            # split full_key by first occurrence of __, if contains __
+            # "key_without_dblunderscore" -> "key_without_dbl_underscore", None, None
+            # "key__with__dblunderscore" -> "key", "__", "with__dblunderscore"
+            key, delim, sub_key = full_key.partition("__")
+            # if key not recognized, remember for suffix matching
+            if key not in valid_params:
+                unmatched_keys += [key]
+            # if full_key contained __, collect suffix for component set_params
+            elif delim:
+                nested_params[key][sub_key] = value
+            # if key is found and did not contain __, set self.key to the value
+            else:
+                setattr(self, key, value)
+                valid_params[key] = value
 
-            # all matched params have now been set
-            # reset object to clean post-init state with those params
+        # all matched params have now been set
+        # reset object to clean post-init state with those params
+        try:
             self.reset()
-        except Exception:
+        except Exception as e:
+            # restore the pre-call state, then report what happened,
+            # the original exception is chained via "from e"
             self.__dict__.clear()
             self.__dict__.update(prev_state)
-            raise
+            raise RuntimeError(
+                f"Error in {type(self).__name__}.set_params, the parameter values "
+                f"passed were rejected when re-running __init__, which raised "
+                f"{type(e).__name__}: {e}. The object has been restored to its "
+                f"state before the set_params call."
+            ) from e
 
         # recurse in components
         for key, sub_params in nested_params.items():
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
Thanks for the contribution. I think this is potentially something to add, but the current implementation has major issues.

* storing the current state can multiply memory requirements. Imagine expensive model weights, which have to exist in two copies.
* I think "silent failure" is the wrong handling of this case. I think there should be an explicit exception at the end.

Further:

* I to not think the entire code needs to be in the try/except loop - only `reset`.
* a clear exception message is missing, the user will not know what has happened

OfficialAbhinavSingh:
Thanks for the review, taking the points in order.

**Only `reset` in the try/except** — agreed. The snapshot stays above the `setattr` loop, since those writes are what leave the object in the bad state, but the `try` now wraps `reset` alone. Side benefit: this drops the re-indentation of the loop, so the change in `_base.py` is `+21/-1`, with `self.reset()` the only existing line touched.

**Memory** — the snapshot is `self.__dict__.copy()`, a shallow dict copy. It stores references to the attribute values, it does not copy the values, so model weights are not duplicated. On an object holding a 128 MB array attribute, peak allocation for an entire failed `set_params` call is 7632 bytes, and `obj.weights is w` is `True` after the rollback.

**"Silent failure" / missing message** — the previous version did re-raise (bare `raise`), so the exception was not swallowed, but you are right that nothing told the user a rollback had happened. It now raises:

```
Error in Est.set_params, the parameter values passed were rejected when
re-running __init__, which raised ValueError: alpha must be positive.
The object has been restored to its state before the set_params call.
```

chained `from` the original, so the `__init__` traceback is preserved.

One point for your call: `test_get_params_after_set_params` catches `(TypeError, ValueError)`, as does sklearn's `check_set_params`. A `RuntimeError` escapes both. I can keep the original exception type instead, if you would rather those checks keep seeing what they see today.

Rationale for rollback vs message-only is in sktime#10695.

OfficialAbhinavSingh:
Went with keeping the original exception type, in a1200bb, so `raise type(e)(...)` rather than `RuntimeError`. The message and the `from e` chaining are unchanged, and `check_set_params` and `test_get_params_after_set_params` keep catching it.

Worth noting the skbase suite did not catch this either way, since the fixture classes do not validate in `__init__`, so the except branch never fires. It shows up on an object that does validate: with `RuntimeError` the error escapes a `(TypeError, ValueError)` catch, with the original type it does not.

1618 passed locally.

